PR #20, opened 2019-11-20
[WIP] Parameter type serialization + Debugging Runtime

insipx:
Debugging the runtime (getting to the bottom of why Registrar.0 fails to decode)

In order to avoid just committing SCALE-encoded byte-arrays to the 'parameters' column of the SQL DB, values that can be serialized are serialized into JSON and then committed to JSONB under 'parameters' column. Values that do not implement `Serialize` are still committed encoded in the JSON object. However, this makes it much more clear for the user of the DB which values need to be decoded, and their exact types (which can be looked up in the corresponding paint module in Polkadot/Substrate)

this branch also includes migration to the re-org of substrate/polkadot


polkadot-archive/src/main.rs
```diff
@@ -20,24 +20,27 @@ use log::{warn, error};
 use failure::Error;
 // use substrate_archive::prelude::*;
 use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
 use substrate_archive::{
     Archive, System, Module, RawExtrinsic,
     OldExtrinsic, ToDatabaseExtrinsic,
     ExtractCall, SrmlExt, NotHandled,
```
4meta5: should `SrmlExt` be renamed at some point to `PaintExt`

insipx: Eventually, Pallet/Pallette now I guess

PR #56, opened 2020-06-01
Experimental Trie Backend

insipx:
Benchmarks, Unit tests, and Experimentation with Parity's TrieDB. This Trie Backend would use TrieDB directly, without going through the substrate client. 

## Justification

Substrate Client is a system designed for use in Full Nodes or Light Clients, and optimized for this. Substrate archive does not require much of the functionality provided by the client, since substrate-archive is essentially a custom layer over RocksDB. This PR attempts to remove layers of abstraction that add computational complexity to indexing substrate-based chains.

Indexing substrate-chains through a substrate-client goes through a few caches that each create their own RwLocks<> on underlying datatypes. Often, the cache itself will be surrounded in a RwLock, as well as the underlying generic that wraps RocksDB. Most of the time, this doesn't pose a problem for reads. There are a few cases, however, where a Cache Object (Such as that found in `CacheDbSync` used by substrates `sc-client-db` changes-tries implementation) makes synchronous blocking calls to RwLock::write() to read from a database or cache. The idea for experimentation is that the caching traditionally done by substrate with an LRU cache serves little purpose for an archive node, since substrate-archive rarely (if ever) will ask for the same block or storage entry twice.

Using the client with caches may hinder performance when attempting to concurrently read for storage changes, I.E., using `query_storage` in multiple threads or from different `futures` contexts' will trigger RwLocks in underlying cache-structures.

This will also allow the removal of a few dependencies that slow down compile times, namely `polkadot-service` (in `kusama-archive`), `sc-service`, and related crates.


insipx:
Closing, changes for this are now included in #57

PR #68, opened 2020-06-24
use old thread stealer and benchmark perf

insipx:
using rayon seems to have regressed performance

Both rayon and this threadpool use crossbeam::deque in order to schedule tasks between threads. However, rayon stores tasks as closures on the heap, and tasks with this threadpool are just queued blocks to execute. Rayon also seemed to use lots of memory comparatively (however, the memory issue could probably have been fixed with some further tweaking).


insipx:
no discernable difference

PR #79, opened 2020-07-18
Add Node-Template back to Workspace and make it compile

insipx:
Add documentation for setting the max open file limit


2075:
ah, was already on it, too 👯

Issue #88, opened 2020-08-31
Raise the fdlimit if we detect `too many open files`


insipx:
This crate: https://github.com/paritytech/fdlimit could help quell the common issue users run into when starting substrate archive for the first time, the `too many open files` error. This error that occurs as a result of the secondary rocksdb database requiring all files of the primary database to be kept open.

liuchengxu:
Sounds like we also want to raise a warning like https://github.com/paritytech/substrate/pull/6956?

insipx:
yeah, a warning with some context would be ideal

liuchengxu:
@insipx Hi, please take a look at #94, I simply borrowed https://github.com/paritytech/substrate/pull/6956, let me know if you want to display more context.

PR #94, opened 2020-09-12
Raise a low fd resources warning message

liuchengxu:
Close https://github.com/paritytech/substrate-archive/issues/88

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -37,6 +37,9 @@ use std::{marker::PhantomData, path::PathBuf, sync::Arc};
 const CHAIN_DATA_VAR: &str = "CHAIN_DATA_DB";
 const POSTGRES_VAR: &str = "DATABASE_URL";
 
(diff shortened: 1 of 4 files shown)


dvdplm (approved):
lgtm

src/archive.rs
```diff
@@ -226,6 +229,16 @@ where
             rt.spec_version,
             last_finalized_block
         );
+        if let Some(new_limit) = fdlimit::raise_fd_limit() {
```
dvdplm, line 232: On windows `raise_fd_limit()` is always `None` which imo is worth mentioning in the docs somewhere, likely the README. Do you mind adding that in this PR?

liuchengxu: @dvdplm Done in https://github.com/paritytech/substrate-archive/pull/94/commits/757ffa95e913a03697a9157bd5db7e7870e99aec. Let me know if you want to put that into another section or something :(

PR #95, opened 2020-09-29
Add configurable `max_block_load` param to limit count of blocks loaded & inserted at a time

### LINKED ISSUE #93: Allow a `max_block_load` parameter to limit the amount of blocks that can be loaded/inserted at any one time
Currently when substrate-archive starts syncing with a chain that is far ahead, it will load as many blocks as it is missing in it's Postgres Database. This could be anything from a few thousand to a couple million blocks. This quickly spikes memory usage if many hundreds of thousands of blocks are loaded at once. Allowing the archive to be configured with a maximum amount of blocks it can load at once would help mitigate  this issue. Currently, this happens in the `blocks` actor:

Dataflow: https://camo.githubusercontent.com/48a4e5bca346a1487ea36cc0e72dca7a394417a7/68747470733a2f2f692e696d6775722e636f6d2f53304e727842332e706e67

Code:
https://github.com/paritytech/substrate-archive/blob/7d3757303f89e498053dd1dd32d1551f5cec173c/src/actors/workers/blocks.rs#L68
https://github.com/paritytech/substrate-archive/blob/7d3757303f89e498053dd1dd32d1551f5cec173c/src/actors/workers/blocks.rs#L91

This may require some refactoring of how the blocks actor indexes blocks, but should be fairly straightforward

emostov:
Closes #93. Refer to issue for feature description.

In order to verify that the refactors in this commit reduce memory usage I conducted some rough benchmarking on my development machine (Intel(R) Xeon(R) E-2288G CPU @ 3.70GHz; 31gb memory)

I first took did a baseline with just a ` ./target/release/polkadot  --chain=polkadot --pruning=archive` node running, which showed the machine stable at 6.34gb of memory usage:
Screen Shot 2020-09-28 at 8 22 16 PM

With both my branch and master I used the following protocol:
- Checkout branch
- Delete database
- Setup database and source the url to shell
- Build archive with `cargo build --release`
- Run archive with  `cargo run --release --  -c test_conf.toml --chain=polkadot`
- Wait 1 minute and then record memory usage

Memory usage for this PR's branch after 1 minute: 7.98gb
Screen Shot 2020-09-28 at 8 26 49 PM

Memory usage for master branch after 1 minute: 9.99gb
master-1min

In sum, after running the aforementioned protocol we see memory usage increase by 1.64gb (7.98 - 6.34) with this PR's branch, while the master branch sees a memory usage increase of 3.5gb (9.99 - 6.34). In other words, the memory usage spike caused by this PR's branch is 53% less than master.

It is important to note that since this testing starts from a fresh database it does not have to deal with "missing" blocks that `re_index` is meant to find and insert.


src/actors/workers/blocks.rs
```diff
@@ -85,27 +91,57 @@ where
...
         } else {
             // a `None` means that the blocks table is not populated yet
-            return Ok(None);
+            return Ok(());
         };
-        let blocks = self.collect_blocks(move |n| numbers.contains(&n)).await?;
-        Ok(Some(blocks))
+
+        if numbers.len() as u32 > self.max_block_load {
+            let mut tmp_set = HashSet::<u32>::with_capacity(self.max_block_load as usize);
```
dvdplm, line 119: I wonder if we could switch to returning a `BTreeSet` from `missing_blocks_min_max` and use `split_off` on the result? Or even limit `missing_blocks_min_max` to `max_block_load` to fetch the batch of blocks we need?

src/actors/workers/blocks.rs
```diff
@@ -85,27 +91,57 @@ where
...
+            .send(BatchBlock::new(self.collect_blocks(fun).await?))
+            .await?;
+        Ok(())
+    }
+
     /// First run of indexing
     /// gets any blocks that are missing from database and indexes those.
     /// sets the `last_max` value.
-    async fn re_index(&mut self) -> Result<Option<Vec<Block<B>>>> {
+    async fn re_index(&mut self) -> Result<()> {
```
dvdplm, line 106: Would it be useful to communicate back to the caller if we're done vs there are batches left to process?

dvdplm (commented):
I'll leave it to Andrew to do a proper review but left a few comments and questions.

src/actors/workers/blocks.rs
```diff
@@ -85,27 +91,57 @@ where
...
+            return Ok(());
         };
-        let blocks = self.collect_blocks(move |n| numbers.contains(&n)).await?;
-        Ok(Some(blocks))
+
+        if numbers.len() as u32 > self.max_block_load {
+            let mut tmp_set = HashSet::<u32>::with_capacity(self.max_block_load as usize);
+            for height in numbers.drain() {
+                tmp_set.insert(height);
+                tmp_set = if tmp_set.len() as u32 > self.max_block_load  {
```
insipx, line 122: What's the rationale behind using HashSet here? It would filter out duplicates however we shouldn't get duplicates unless something is wrong with our SQL query

Rather than use drain/HashSet here, we could iterate using the `chunks` method defined on slices and check if the number is contained within that chunk since we don't need to worry about passing around an owned `height`. This should remove the need to instantiate a new HashSet on every iteration

src/actors/workers/blocks.rs
emostov: From my understanding, when `re_index` returns it should have completely caught up until `last_max` and thus there will not be any batches left to process.

src/actors/workers/blocks.rs
emostov: About to try this out, but I would think `chunks().into_iter()` would return owned chunks, and from my understanding this consumes `chunks()` and just change the way its referenced but not touch the heap. 

Also I was trying to figure out yesterday, does `to_vec()` just changed what is stored on the stack or does it also affect the heap?

emostov: I ended up trying out a solution that modifies the SQL query. From what I read, using LIMIT is said to reduce memory usage of the query and SQL can do some optimizations under the hood. As I am typing right now realizing though that I haven't checked if awaiting the metadata actor is enough to make sure the current batch is inserted - if not this code is not correct.

I tried out solutions with BTreeSet and Vec as well, but this SQL query might be more space and time efficient than the Vec solution by some constants. (I couldn't figure out a a BtreeSet solution with better than nlogn time)

My main issue right now is I can't figure out how to test the modified query because I can't get the database to have missing entries. I kept on trying to just kill the processes fro archive while syncing, but on restart the missing blocks query always comes back empty. Is there an easy way I could seed a DB with just one block at high height?

dvdplm: > I was trying to figure out yesterday, does `to_vec()` just change what is stored on the stack or does it also affect the heap?

I am pretty sure `to_vec()` reallocates and clone the data in this case.

dvdplm: > I ended up trying out a solution that modifies the SQL query.

Sounds like the most straight forward approach.

> Is there an easy way I could seed a DB with just one block at high height?

Why not just insert a new record? Clone a row of data and manually edit the height column?

src/actors/workers/blocks.rs
dvdplm: ofc, my bad. 👍

src/actors/workers/blocks.rs
dvdplm: @emostov just a sidenote, I've sometimes used [`alloc_counter`](https://lib.rs/crates/alloc_counter) when I had doubt about how much a piece of code allocates. It's far from perfect as it can't see any allocations that happens outside the Rust code (e.g. Postgres or Rocksdb) but still handy at times, e.g. in benchmarks or while debugging something.

insipx: > Is there an easy way I could seed a DB with just one block at high height?

If you install the SQLx CLI: https://github.com/launchbadge/sqlx/tree/master/sqlx-cli you should be able to drop the current database  and create a new one, then run the migrations for substrate-archive with `cargo sqlx migrate run` after setting the DATABASE_URL environment variable.

Then, this query should insert block 101000 (should work through psql):
```sql
INSERT INTO blocks (parent_hash, hash, block_num, state_root, extrinsics_root, digest, ext, spec) VALUES (E'\\x8da3783495634a1645cddb0d5b72743170971a97b68391ce2eb4cbe6a1abebcc', E'\\xa8e323dce4492818f6ae828ea0e483c572536ec8b56a742c351b25fccc0c428f', 101000, E'\\x407c1a180a91c3fed506ac7a1695b31bc830e46c98a8d8397430f2f0bc8f79b0', E'\\xfebdb8c5a254c0e58831b1d01d398abd74979a5ec468797ef9a3b80e9f5d4fb0', E'\\x08064241424534023b0000001eeea60f0000000005424142450101b406381a5d732324723a54ee0951eecf43b621af610ce844fc297445e270de416311da7ef66300f8d637fc9ea907e19478ebcc9307589a376ba05215ca582d82', E'\\x0c280402000b20df6cd86e011c0409001a2a06001004140000', 1028);
```

> I haven't checked if awaiting the metadata actor is enough to make sure the current batch is inserted - if not this code is not correct.

I just checked up on this and you're right, it won't, but it should be a simple fix. Adding another `.await` here:
https://github.com/paritytech/substrate-archive/blob/70c6d9d3bad8cbfb5ac7ed407eb8acc04fbf26d3/src/actors/workers/metadata.rs#L88

Is all it should take to fix, the first future returns a future that can be used to wait for the result of the first future (not really intuitive, but that api is out-of-scope for this PR I think)

emostov:
Ok - I think I have a working solution with the LIMIT SQL query. 

Using the methods @insipx recommended, I scripting out dropping, creating, migrating, and seeding a kusama database with one block. This was very helpful for testing and I caught some critical issues with my logic.

I did some benchmarking again using a similar protocol as last time, except with the database seeded with a single block.

Baseline memory usage of: 4.50gb
baseline

This PR's memory usage after 1 minute: 5.76gb
zeke-branch

Master branches memory usage after 1 minute: 6.70gb
master

I will note there was some noise on the machine used for testing from other users that may have impacted these numbers.

src/database/queries.rs
```diff
@@ -59,19 +59,24 @@ pub(crate) async fn missing_blocks(conn: &mut PgConnection) -> Result<Vec<u32>>
...
         "SELECT generate_series
         FROM (SELECT $1 as a, max(block_num) as z FROM blocks) x, generate_series(a, z)
         WHERE
         NOT EXISTS(SELECT id FROM blocks WHERE block_num = generate_series)
         ORDER BY generate_series ASC
+        LIMIT $2
         ",
     )
     .bind(min as i32)
+    .bind(max_block_load as i32)
```
dvdplm, line 79: Not a fan of these casts here. If `max_block_load` is bigger than `i32::MAX` [it will wrap around to a negative number](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=057ba7218ff448ee8845be9973cd021c).
Is this something `sqlx` requires us to do or can we use `u32` throughought? If we can't then we should use `try_into` here.

src/database/queries.rs
```diff
@@ -59,19 +59,24 @@ pub(crate) async fn missing_blocks(conn: &mut PgConnection) -> Result<Vec<u32>>
...
     conn: &mut PgConnection,
     min: u32,
+    max_block_load: u32,
 ) -> Result<HashSet<u32>> {
     Ok(sqlx::query_as::<_, (i32,)>(
         "SELECT generate_series
         FROM (SELECT $1 as a, max(block_num) as z FROM blocks) x, generate_series(a, z)
         WHERE
         NOT EXISTS(SELECT id FROM blocks WHERE block_num = generate_series)
         ORDER BY generate_series ASC
```
dvdplm, line 74: Not your code, but I think this query is hard to read. 
`SELECT generate_series` – don't use the same name as the function used
`SELECT $1 as a, max(block_num) as z FROM blocks) x, generate_series(a, z)` - don't use single letter variable names

Overall I think it's good practice to use in-line comments to explain what each sub-query does.

Also, do we really need to generate a series here? Isn't it enough to figure out what the interval is and return a [`Range`](https://doc.rust-lang.org/std/ops/struct.Range.html)?

src/actors.rs
```diff
@@ -241,9 +246,13 @@ where
...
             .spawn();
-        let blocks =
-            workers::BlocksIndexer::new(ctx.backend().clone(), db_pool.clone(), metadata.clone())
-                .spawn();
+        let blocks = workers::BlocksIndexer::new(
+            ctx.backend().clone(),
+            db_pool.clone(),
+            metadata.clone(),
+            ctx.max_block_load,
+        )
```
dvdplm, line 254: Do you think `BlocksIndexer::new` could be changed to take only a `ctx` and `db_pool`? Seems like we're piling on arguments here and that could be done in `new()` itself?

dvdplm:
@insipx Do you have a suggestion for zeke for a more formal benchmark than eyeballing `htop`?

insipx:
> @insipx Do you have a suggestion for zeke for a more formal benchmark than eyeballing `htop`?

Sure,

Not really super-formal but in the past I have mostly used [HeapTrack](https://github.com/KDE/heaptrack) to calculate memory usage. There's also rust-native [memory-profile](https://github.com/koute/memory-profiler) that I've heard some other people have had success with but haven't tried it out myself.

@dvdplm do you think it's worth creating benchmarks for memory usage through something like cargo-bench or otherwise? This hasn't seemed super necessary before, but it might be a good way to standardize tracking how much memory improvement a PR has.

src/database/queries.rs
insipx: will require `try_into`, since there's no direct PostgreSQL equivalent type to an unsigned integer sqlx chooses not to represent it when binding values.

dvdplm:
> do you think it's worth creating benchmarks for memory usage through something like cargo-bench or otherwise?

I think `cargo bench` or criterion would not work so well here. We want "run these couple of processes here for 10 minutes and sample overall memory usage every 10seconds; compare result to the baseline in this file over here" which is significantly harder to set up (and maintain!) than low-level benchmarks. If there is something easy to use to do the sampling part then perhaps @emostov could cook up a bash script to do it, but maybe it's a bit out of scope for this PR. Definitely worth having though imo.

src/database/queries.rs
insipx: I agree, the query itself can be made easier to read (a mistake on my end), ie: 

```sql
SELECT missing_number
FROM (SELECT $1 as min_num, MAX(block_num) AS max_num FROM blocks) minmax, 
    GENERATE_SERIES(min_num, max_num) AS missing_number
WHERE
NOT EXISTS (SELECT id FROM blocks WHERE block_num = missing_number)
ORDER BY missing_number ASC
```

On the interval -- I'm not actually sure how we could accomplish this with Postgres (Postgres can't query things that don't exist) other than not using Postgres and just returning all the block numbers and doing it in rust rather then with a query, or generating the series before doing the query, where the query would return numbers which don't exist from the series we give it. 

I think the Postgres solution is good because it contains the set logic to SQL rather than dealing with it in rust

dvdplm: > I think the Postgres solution is good because it contains the set logic to SQL rather than dealing with it in rust

Totally agree.

My idea for the `Range` was that we could perhaps return just two numbers from the query, first and last in the range, instead of a whole series. I was thinking about wrapping it in another `SELECT` with the current query as a subquery, but maybe that doesn't work? The idea was: it's a bit annoying we need a to shuffle a whole `HashSet` around when all we need to know (?) is what the extremities of the set are.

src/database/queries.rs
emostov: Can we just cast to an i64 (bigint in postgres)?

src/database/queries.rs
emostov: > it's a bit annoying we need a to shuffle a whole HashSet around when all we need to know (?) is what the extremities of the set are.

If I understand the query correctly, we are not just getting a straight series of numbers and throwing it into a `HashSet`. Instead we first generate the straight series, and then return the numbers from the series that are not in the blocks table. For extremities of the series we only need the query MAX(block_num) since we already feed in the min.

src/actors.rs
emostov: In order to construct metadata within `BlocksIndexer::new` we would need to make it an async function and copy the code for instantiating the metadata actor, which imo seems a bit heavy handed. I did however remove the `backend` and `max_block_load` param since they are directly on `ctx`. 

I think if I was trying to do this in TS I would pass the same args in as originally, but just use an object for the argument, which can be cleanly de-structured - I don't think there is clean equivalent in rust for de-structuring params like that though. The reason I bring it up is that I think it decouples the function from the interface of `ctx` and imo it is more declarative/ maintains the division of responsibility.

emostov:
> > do you think it's worth creating benchmarks for memory usage through something like cargo-bench or otherwise?
> 
> I think `cargo bench` or criterion would not work so well here. We want "run these couple of processes here for 10 minutes and sample overall memory usage every 10seconds; compare result to the baseline in this file over here" which is significantly harder to set up (and maintain!) than low-level benchmarks. If there is something easy to use to do the sampling part then perhaps @emostov could cook up a bash script to do it, but maybe it's a bit out of scope for this PR. Definitely worth having though imo.

Do we want memory profiling prior to this PR being merged? Would also be good to clarify goals of memory profiling. From what it seems like to me this an edge case of archive since the changes here will likely just affect the initial sync. More useful from memory profiling might be trying to optimize the value of `max_block_load`, but I imagine that is resource dependent.

src/actors/workers/blocks.rs
```diff
@@ -85,27 +86,55 @@ where
...
+            }
+        }
+
+        self.last_max = cur_max;
+        log::info!("{} missing blocks", missing_blocks);
+
+        Ok(())
     }
 
+    /// Crawl up to `max_block_load` blocks that are greater then the last max
```
dvdplm, line 131:
```suggestion
    /// Crawl up to `max_block_load` blocks that are greater than the last max
```

dvdplm (approved):
lgtm modulo a typo

dvdplm:
> Do we want memory profiling prior to this PR being merged?

No, I don't think that's in scope for this PR. Let's see what running archive in production looks like and re-assess.

dvdplm:
@insipx what's up with the angry linter? Does it come from this PR or was it like this before?

src/actors.rs
insipx: I think sometime in the future we may try to refactor the architecture of `System` and how it spawns/creates actors and connects external services (client/backend) to them, possibly by splitting it up into two separate structs or some other strategy. It's one of the parts of archive that I've struggled  to 'get right' in terms of the API, and never have been quite satisfied with.

insipx (approved):

insipx:
gg

> @insipx what's up with the angry linter? Does it come from this PR or was it like this before?

I think #99 should fix it, I thought #96 would fix it but it did not

src/database/queries.rs
insipx: I believe casting to i64 would work too

PR #96, opened 2020-09-29
QoL pin Substrate to latest revision Polkadot uses

insipx:
This pins substrate to the latest revision that the latest polkadot release is using (0.8.24). Previously, substrate-archive would use the latest master of polkadot/substrate, however this led to some situations where a substrate update would break archive. Tracking the latest revision that Polkadot uses should lead to less compilation conflicts.

The latest polkadot release revision was chosen as not to break polkadot-archive. Since the runtime is imported from polkadot, mismatched versions of substrate would cause compilation to complain that traits aren't implemented.

- includes a commit that runs rustfmt on the codebase
- removes `jsonrpsee`dependency and the old `rpc.rs` file containing functions that are no longer used


Cargo.toml
```diff
@@ -38,28 +38,27 @@ chrono = { version = "0.4", optional = true }
...
-sp-core = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-core" }
-sp-api = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-api" }
-sp-block-builder = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-block-builder" }
-sp-version = {  git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-version" }
-sc-executor = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sc-executor" }
-sc-chain-spec = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sc-chain-spec" }
-sp-trie = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-trie" }
-sp-state-machine = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-state-machine" }
-sp-io = { git = "https://github.com/paritytech/substrate", branch = "master", package = "sp-io" }
+sp-database = { git = "https://github.com/paritytech/substrate", package="sp-database" , rev = "647ad15565d7c35ecf00b73b12cccad9858780b9" }
```
emostov, line 47: Outside of the scope of this PR, any idea i the new substrate versioning allow us to use `tag` instead of `rev` in the future?

emostov (approved):
Looks clear cut and sensible. For context for myself, would desub alleviate the issues of substrate deps complaining about undefined types coming from the runtime? If there was ever a breaking change to a type in substrate (i.e its not appended but just totally redefined), would we need to keep different versions around of substrate deps?

Cargo.toml
dvdplm: I hope so, or even better, just straight up crate versions. :)

I believe this PR is rather temporary though, as of the next kusama/polkadot releases we should be able to use the `2.0.0` releases.

bin/polkadot-archive/Cargo.toml
```diff
@@ -8,8 +8,8 @@ edition = "2018"
 
 [dependencies]
 substrate-archive = { path = "../../", features = ["logging"] }
-polkadot-service = { package = "polkadot-service", git = "https://github.com/paritytech/polkadot", branch = "master" }
-sc-chain-spec = { package = "sc-chain-spec", git = "https://github.com/paritytech/substrate", branch = "master" }
+polkadot-service = { package = "polkadot-service", git = "https://github.com/insipx/polkadot", branch = "substrate-archive" }
```
dvdplm, line 11: This one is special it seems. I think it merits a comment explaining why `archive` needs its own fork and what needs to happen upstream to let us switch to a released version.

dvdplm (approved):
Minor nitpicks.

q: do we run `cargo fmt` in CI? If not, let's add it!

insipx:
> Minor nitpicks.
> 
> q: do we run `cargo fmt` in CI? If not, let's add it!
It's been running but not actually formatting any of the code, just updated GH Actions to do so, though

bin/polkadot-archive/Cargo.toml
insipx: Added a comment, but this happens since cargo ignores the lockfile. If upstream polkadot starts using substrate versions instead of master branch in Cargo.toml, then we could switch to do that too. Otherwise, changing the pinned version in Cargo.toml is the only way to pin the version universally. Not sure if the plan is to switch to actual versions when polkadot updates to substrate 2.0

PR #101, opened 2020-10-08
Refactor SQL queries to enable type checking

### LINKED ISSUE #98: Change to type-checked SQL queries with `query` and `query_as` macro
SQLx includes compile-time verified queries with `query!()` and `query_as!()`. It would be good to refactor any queries that currently use the `sqlx::query` methods to the compile-time-verified variants. This will give us assurance that the queries are semantically correct, as well as work correctly with the db schema at compile time. 

[`query!()`](https://docs.rs/sqlx/0.3.5/sqlx/macro.query.html)
[`query_as!()`](https://docs.rs/sqlx/0.3.5/sqlx/macro.query_as.html)

emostov:
Closes #98

Changes:
- ~IMPORTANT: In order to compile `archive` the `DATABASE_URL` environment variable **must** be set to a database with the relevant tables at build time (see [`sqlx` docs](https://docs.rs/sqlx/0.3.5/sqlx/macro.query.html#requirements)).~ This should not be an issue now as I checked in the artifact from `cargo sqlx prepare` and added the `offline` feature

- Replace all queries with `query_as!` macro. `query_as!` requires the return type of `query_as` to always be a struct, so several new structs are introduced. 

- Breaking change for `missing_blocks_stream` - stream item goes from `Result<(i32,)>` to `Result<Series>` where `Series` is a new struct that describes the model of data return by `sqlx`.

Areas for feedback:
- In functions that use `GENERATE_SERIES`, each value within the returned series is an `Option`. As far as I can tell these will always be `Some`, but I am not certain because I could not find anything in the `sqlx` docs that elaborated on the matter
- ~Is there a way to setup a database and point `DATABASE_URL` to it for the CI checks?~ None issue now due to offline mode, but as @insipx mentioned, this may be a worthwhile endeavor in the future to support unit tests.

Other notes:
- I tried using i64s to query so as to avoid i32 from u32 overflow, but the i64 is rejected at the postgres level


insipx:
> Is there a way to setup a database and point `DATABASE_URL` to it for the CI checks?

Still have to look at the rest of this, but yeah this is going to complicate CI a little bit

I've setup a database for rust project before here: https://github.com/insipx/coil/blob/master/.github/workflows/coverage.yml
and run unit tests against it. In addition to that we'd need to somehow install the sqlx-cli command with GH Actions and run the migrations before running the check phase

Another options is to do this via committing the result of `cargo sqlx prepare` which will allow compilation to go ahead without a `database_url` necessarily being set: https://github.com/launchbadge/sqlx/tree/master/sqlx-cli


With a Postgres database we could potentially devise a unit-testing strategy for queries as well

src/database/queries.rs
```diff
@@ -117,84 +154,110 @@ pub(crate) async fn blocks_storage_intersection(
...
 ) -> Result<Vec<u32>> {
-    let query = String::from("SELECT block_num FROM blocks WHERE block_num = ANY ($1)");
-    let row = sqlx::query_as::<_, (i32,)>(query.as_str())
-        .bind(nums)
-        .fetch_all(conn)
-        .await?;
-    Ok(row.into_iter().map(|r| r.0 as u32).collect())
+    let safe_nums: Vec<i32> = nums
+        .iter()
+        .map(|n| i32::try_from(*n).unwrap_or(i32::MAX))
```
dvdplm, line 231: You should be able to `filter` or `filter_map` here I think, to avoid keeping bad input as `i32::MAX`.

src/lib.rs
```diff
@@ -172,9 +172,9 @@ mod test {
                     TRUNCATE TABLE metadata CASCADE;
                     TRUNCATE TABLE storage CASCADE;
                     TRUNCATE TABLE blocks CASCADE;
-                    TRUNCATE TABLE frame_system CASCADE;
                     TRUNCATE TABLE _background_tasks
                     ",
+                    // TRUNCATE TABLE frame_system CASCADE;
```
dvdplm, line 177: cleanup

src/database/queries.rs
```diff
@@ -25,76 +25,112 @@ use sp_runtime::traits::Block as BlockT;
...
+        ORDER BY missing_num ASC
         ",
     )
     .fetch(conn)
     .map_err(Error::from)
 }
 
-/// get missing blocks from relational database
+/// Get the `block_num` of missing blocks from relational database
 #[allow(unused)]
```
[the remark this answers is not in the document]
insipx: We can just remove these methods, they're holdovers that I kept for reference in case I wanted to use a stream with one of the queries while testing using a stream vs returning as a vec, but are not used for anything currently.

src/database/queries.rs
```diff
@@ -117,84 +154,110 @@ pub(crate) async fn blocks_storage_intersection(
...
         .fetch_all(conn)
-        .await?;
-    Ok(rows.into_iter().map(|r| r.0 as u32).collect())
+        .await?
+        .into_iter()
+        .map(|r| r.version as u32)
+        .collect())
 }
 
+/// Get all the blocks in `_background_tasks` WHERE job_type = 'execute_block'
```
insipx, line 255:
```suggestion
/// Get all full blocks queued for execution in the background task queue.
```

src/database/queries.rs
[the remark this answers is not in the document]
emostov: Removed

emostov:
Updated with feedback. Noteworthy change is that `check_if_meta_exists` now returns `Ok(false)` if the conversion to `i32` overflows. 

One other thing worth noting is that if you just want to compile/check without having a database, you might have to set environment variable `SQLX_OFFLINE=true`. Should we put a note about that in the wiki?

insipx:
> Updated with feedback. Noteworthy change is that `check_if_meta_exists` now returns `Ok(false)` if the conversion to `i32` overflows.
> 
> One other thing worth noting is that if you just want to compile/check without having a database, you might have to set environment variable `SQLX_OFFLINE=true`. Should we put a note about that in the wiki?

Yeah a quick note in the `requirements` section would be nice!

insipx (approved):

dvdplm:
> > One other thing worth noting is that if you just want to compile/check without having a database, you might have to set environment variable `SQLX_OFFLINE=true`. Should we put a note about that in the wiki?

Can we make that the default? Or alternatively, check the env for it and use [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) to give users a chance to figure out what's going on without reading docs (or open tickets, which is their most likely course of action)?

emostov:
> > > One other thing worth noting is that if you just want to compile/check without having a database, you might have to set environment variable `SQLX_OFFLINE=true`. Should we put a note about that in the wiki?
> 
> Can we make that the default? Or alternatively, check the env for a database connection and use [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) to give users a chance to figure out what's going on without reading docs (or open tickets, which is their most likely course of action)?

So right here https://github.com/paritytech/substrate-archive/blob/a1ce3674700d77b1b16595a6fc819c12eac3230a/src/archive.rs#L152
we give an error when someone starts archive to use it as a program but did not feed in a url in the config or as an env. The situation where you don't get that error though is when you are just compiling the source the code but not running it (`cargo build`). In that case the compiler complains that `DATABASE_URL` is not set but it does not mention that `SQLX_OFFLINE=true` can be set. I am not sure how to build in a compile error that would only fire when building but not when it is being run as a program.

dvdplm:
> The situation where you don't get that error though is when you are just compiling the source the code but not running it (`cargo build`). In that case the compiler complains that `DATABASE_URL` is not set but it does not mention that `SQLX_OFFLINE=true` can be set. I am not sure how to build in a compile error that would only fire when building but not when it is being run as a program.

Ok, I'll see if I can make a PR; no reason to hold up this one here.

dvdplm (approved):

dvdplm:
> Ok, I'll see if I can make a PR; no reason to hold up this one here.

This can't be done with `compile_error!`, my bad. I can't repeat the error I saw last time I tried so not sure what I was rambling about, sorry for the noise.

PR #102, opened 2020-10-10
Clean up clippy warnings/errors

### LINKED ISSUE #99: Clean up clippy warnings and errors
Right now the CI check `Rust / Lints` always fails on historical code due to clippy complaining. In order to expunge those errors from the CI check and allow meaningful use of clippy in development, those warnings should be silenced or the code should be corrected.

emostov:
Closes #99, see issue for fix description.

### DIFF
--- a/src/actors.rs
+++ b/src/actors.rs
@@ -153,7 +153,7 @@ where
         max_block_load: u32,
     ) -> Result<Self> {
         let context = ActorContext::new(
-            backend.clone(),
+            backend,
             client_api.clone(),
             workers,
             pg_url.to_string(),
@@ -298,7 +298,7 @@ where
             .collect();
         let difference: HashSet<u32> = missing_storage_nums
             .difference(&blocks)
-            .map(|b| *b)
+            .copied()
             .collect();
         missing_storage_blocks.retain(|b| difference.contains(&(b.block_num as u32)));
         let jobs: Vec<crate::tasks::execute_block::Job<B, R, C>> =
--- a/src/actors.rs
+++ b/src/actors.rs
@@ -292,11 +292,10 @@ where
             .map(|b| Ok((*b?.header().number()).into()))
             .collect::<Result<_>>()?;
         let mut missing_storage_blocks = queries::blocks_storage_intersection(conn).await?;
-        let missing_storage_nums: HashSet<u32> = missing_storage_blocks
+        let difference: HashSet<u32> = missing_storage_blocks
             .iter()
             .map(|b| b.block_num as u32)
-            .collect();
-        let difference: HashSet<u32> = missing_storage_nums
+            .collect::<HashSet<u32>>()
             .difference(&blocks)
             .copied()
             .collect();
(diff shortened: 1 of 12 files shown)


src/actors.rs
```diff
@@ -298,7 +298,7 @@ where
             .collect();
         let difference: HashSet<u32> = missing_storage_nums
             .difference(&blocks)
-            .map(|b| *b)
+            .copied()
```
dvdplm, line 301: This change is good. I wonder however if we couldn't combine `missing_storage_nums` and this pile of combinators into one, thus avoiding having to allocate for `missing_storage_nums`?

src/actors/workers/database.rs
```diff
@@ -96,7 +96,7 @@ impl<B: BlockT> DatabaseActor<B> {
     async fn batch_storage_handler(&self, storage: Vec<Storage<B>>) -> Result<()> {
         let mut conn = self.db.conn().await?;
         let mut block_nums: Vec<u32> = storage.iter().map(|s| s.block_num()).collect();
-        block_nums.sort();
+        block_nums.sort_unstable();
```
dvdplm, line 99: What did clippy complain about here?

src/backend/util.rs
```diff
@@ -141,9 +141,9 @@ pub fn read_db<Block>(
 where
     Block: BlockT,
 {
-    block_id_to_lookup_key(db, col_index, id).and_then(|key| match key {
-        Some(key) => Ok(db.get(col, key.as_ref())),
-        None => Ok(None),
+    block_id_to_lookup_key(db, col_index, id).map(|key| match key {
+        Some(key) => db.get(col, key.as_ref()),
+        None => None,
```
dvdplm, line 146: Would this work?

```
    block_id_to_lookup_key(db, col_index, id)
        .map(|key|
            key.and_then(|key| db.get(col, key.as_ref()))
        )
```

If it does I'm still not sure I like it better. Explicit is good sometimes.

dvdplm (approved):
lgtm modulo nits

src/actors/workers/database.rs
insipx: I think it's this lint: https://github.com/rust-lang/rust-clippy/issues/5762

insipx (approved):

src/actors/workers/database.rs
dvdplm: I see, TIL. :)

src/actors.rs
emostov: Ok, I just combined it into one, however I don't see how this avoids a non-constant allocation. From my understanding we would still have to allocate at `.collect::<HashSet<u32>>()`, which is the equivalent of `missing_storage_nums` minus the variable binding on the stack. But I am guessing I am missing something?

dvdplm: I don't know how what you have now optimizes, but before we had two `collect()` calls (which each allocates); with this last change I'd hope we only allocate once and for a smaller collection, i.e. just the difference.

dvdplm (approved):

PR #103, opened 2020-10-13
Fix a few warnings

dvdplm:
Fix warnings and remove whitespace at the end of one of the query lines.

### DIFF
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -334,8 +334,8 @@
       ]
     }
   },
-  "f1a15dd2b69242061774eb65c073174b671182a43ca6839246f5b6e6039ae46b": {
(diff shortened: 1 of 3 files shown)


insipx (approved):

emostov (approved):
I don't understand where the "ref` on an entire `let` pattern is discouraged, take a reference with `&` instead" lint warning is coming from. Otherwise looks good to me. Also curious, what command did you run to see these warnings? They didn't show up for me when I was doing the clippy clean up.

dvdplm:
>  don't understand where the "ref`on an entire`let`pattern is discouraged, take a reference with`&` instead" lint warning is coming from.

It's because CI runs stable, so you'll see them if you run `cargo +stable clippy -- -D warnings`. Not sure where it comes from either but somewhere in the macro I think?

> Also curious, what command did you run to see these warnings?

`cargo test`/`cargo check --tests`

PR #106, opened 2020-10-21
do not print postgres url in plaintext

insipx:
Pretty self-explanatory -- this could be problematic in some environments from a security standpoint. Ideally we could just print the name of the database to logs when we do migrations, but this PR is just a one-line change that removes the line printing the full URL which includes the database password.


insipx:
Since this is just a one-line change that only effects logs, I'm just going to merge this w/o review

dvdplm (approved):
Maybe we could leave just `Running migrations` but this is fine too.

niklasad1 (approved):

PR #108, opened 2020-10-25
Refactor: workspace with core, common, and backend; Make blockchain DB generic

### LINKED ISSUE #89: Encapsulate the substrate backend
`src/backend/` should be moved into it's own crate. This crate should only expose the `Client` and `ReadOnlyBackend` that substrate-archive can use to work with.

 `ReadOnlyDatabase` should be abstracted into an archive-specific trait that can be implemented for different database types. rel #76. This trait can then be included in `ReadOnlyBackend` instead of the currently-included struct. This would make the backend-database implementation opaque and generic to substrate-archive, allowing users to use whichever database they like.

The trait should: 
- read key/value pairs of the database (`get`)
- iterating over all blocks in the database (`iter`)
- catch up with the latest information added to the database (`catch_up_with_primary`)

Look in https://github.com/paritytech/substrate-archive/blob/70c6d9d3bad8cbfb5ac7ed407eb8acc04fbf26d3/src/backend/database.rs#L48 for the specific functionality required. This trait should replace the struct in `ReadOnlyBackend` defined here:
https://github.com/paritytech/substrate-archive/blob/70c6d9d3bad8cbfb5ac7ed407eb8acc04fbf26d3/src/backend/read_only_backend.rs#L49

This should also encapsulate out many of the substrate dependencies when working on stuff related to the SQL database/actors. The 'backend' crate (or whatever it will be named) should also expose a way of getting the Archive Client implementing the `ApiAccess` trait and related types (This `ApiAccess` trait should also be moved into the new `backend` crate). This is the function that does this now: https://github.com/paritytech/substrate-archive/blob/2f3b4ec766ae53a60b76c1b78cd384bc00bdae91/src/backend/frontend.rs#L44

This could be split into two PR's, one for ParityDB support and the abstract ReadOnlyDatabase trait, and one for creating a separate crate for `backend`.

### LINKED ISSUE #115: Try using static generics instead of dynamic dispatch for ReadOnlyBackend
it might be worth using static generics here:

https://github.com/paritytech/substrate-archive/pull/108/files/1fc2c7c47b230b43510de8fef82967ae02bbaa5a#diff-26082dae258bbbb4883f22123d203832ebce121a5ab81a999f8288be84c323bbR48

Dynamic dispatch means jumping through vtables and it could be more performant to use static generics instead

emostov:
Closes #89.
closes #115

### `substrate-archive` in 3 crates:
1) `substrate-archive`: this crate holds the majority of the code, excluding the blockchain database implementation and some types that are common to the blockchain database implementation and `substrate-archive`
2) `substrate-archive-common`: this crate has some miscellaneous enum/struct/type/trait definitions and implementations that are common to both `substrate-archive` and `substrate-archive-backend`
3) `substrate-archive-backend`: holds the blockchain database implementation

As per offline discussion with @insipx, the exports of `substrate-archive-backend` are not trimmed down to just `Client` and `ReadOnlyBackend` as originally stated in the issue. We leave public a wider array of exports instead of cramming them all into `Client`, which would then get un-necessarily messy with types. Instead the exports reflect that specific users use different aspects of the backend and should just utilize the components they need.

This PR does not address ParityDB support, as mentioned in #89.

### Other changes:
- Silence `clippy` warnings from query.rs (upstream warnings come from launchbadge/sqlx#594)
- Use static generics for ReadOnlyBackend (#115)

### Directly related future work:
- Support ParityDB (#76)
- #123


substrate-archive-backend/src/lib.rs
```diff
@@ -18,25 +18,31 @@
...
+// re-exports
+pub use self::{
+    block_exec::BlockExecutor,
+    frontend::{runtime_api, TArchiveClient},
+    read_only_backend::ReadOnlyBackend,
+    runtime_version_cache::RuntimeVersionCache,
+    util::open_database,
+};
+
+#[cfg(feature = "logging")]
```
dvdplm, line 43: When is it useful to run without logging?

substrate-archive-backend/src/runtime_version_cache.rs
```diff
@@ -87,7 +87,7 @@ impl<B: BlockT> RuntimeVersionCache<B> {
             .storage(hash, well_known_keys::CODE)
             .ok_or_else(|| Error::from("storage does not exist"))?;
 
-        let code_hash = crate::util::make_hash(&code);
+        let code_hash = substrate_archive_common::util::make_hash(&code);
```
dvdplm, line 90: Import and call `util::make_hash()`

substrate-archive-common/src/database.rs
```diff
@@ -0,0 +1,13 @@
...
+pub type KeyValuePair = (Box<[u8]>, Box<[u8]>);
+
+// Archive specific K/V database reader implementation
+pub trait ReadOnlyDatabaseTrait: Send + Sync {
+    /// Read key/value pairs from the database
+    fn get(&self, col: u32, key: &[u8]) -> Option<Vec<u8>>;
+    /// Iterate over all blocks in the database
+    fn iter<'a>(&'a self, col: u32) -> Box<dyn Iterator<Item = KeyValuePair> + 'a>;
+    /// Catch up with the latest information added to the database
+    fn try_catch_up_with_primary(&self) -> Result<()>;
```
dvdplm, line 12: Maybe we can ditch the `try_` here?

substrate-archive-common/src/database.rs
```diff
@@ -0,0 +1,13 @@
...
+
+pub type KeyValuePair = (Box<[u8]>, Box<[u8]>);
+
+// Archive specific K/V database reader implementation
+pub trait ReadOnlyDatabaseTrait: Send + Sync {
+    /// Read key/value pairs from the database
+    fn get(&self, col: u32, key: &[u8]) -> Option<Vec<u8>>;
+    /// Iterate over all blocks in the database
+    fn iter<'a>(&'a self, col: u32) -> Box<dyn Iterator<Item = KeyValuePair> + 'a>;
+    /// Catch up with the latest information added to the database
```
dvdplm, line 11: Document failure conditions?

substrate-archive-common/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub mod database;
```
dvdplm, line 1: Missing licensing header in this file.

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +104,37 @@ fn format_opt(file: Option<String>) -> String {
         Some(f) => f,
     }
 }
+
+/// get the path to a local substrate directory where we can save data
```
dvdplm, line 108:
```suggestion
/// Get the path to a local substrate directory where we can save data.
```

I think we need more info here. What is "local substrate directory" (for the various platforms)? What if I configured my node to store data elsewhere? If archive makes hard assumptions on the paths we should document that. And also mention that archive's data is in `some/path/substrate_archive`.

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +104,37 @@ fn format_opt(file: Option<String>) -> String {
...
+pub fn substrate_dir() -> PathBuf {
+    if let Some(base_dirs) = dirs::BaseDirs::new() {
+        let mut path = base_dirs.data_local_dir().to_path_buf();
+        path.push("substrate_archive");
+        path
+    } else {
+        panic!("Couldn't establish substrate data local path");
```
dvdplm, line 115: Why panic here but `std::process::exit` in `create_dir`? Maybe the decision on what to do should be in the caller, so that we have a single place where decisions like exit/panic/retry/call mum can be made?

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +104,37 @@ fn format_opt(file: Option<String>) -> String {
...
+    }
+}
+
+/// create an arbitrary directory on disk
+/// panics if it fails because of anything other than the directory already exists
```
dvdplm, line 120: I think this need some re-wording.

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +104,37 @@ fn format_opt(file: Option<String>) -> String {
...
+            _ => {
+                error!("{}", e);
+                std::process::exit(0x0100);
+            }
+        }
+    }
+}
+
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hasher as _;
```
dvdplm, line 134: I prefer `use`s on top.

substrate-archive/Cargo.toml
```diff
@@ -0,0 +1,82 @@
...
+fdlimit = "0.2"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+serde-aux = "0.6.1"
+dirs = { version = "3", package = "directories" }
+sqlx = { version = "0.4.0-beta.1", default-features = false, features = ["postgres", "macros", "runtime-async-std", "migrate", "json", "offline"] }
+async-trait = "0.1"
+hex = "0.4"
+itertools = "0.9"
+# 0.7.2 introduces breaking changes with Send/Sync bounds
```
dvdplm, line 20: Does this mean that `0.7.2` mustn't be used? Best use `flume = "=0.7.1` then, or something similar that stops `0.7.2` to be used.
If upgrading requires a refactor we don't want to do in this PR it would be good to add an issue for it.

dvdplm (commented):
I like where this is going. I think we should merge this sooner rather than later and then iterate on it (if needed).

Maybe we can add `rustfmt` in here as well, as it's a pretty noisy PR anyway? @insipx thoughts?

substrate-archive-backend/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# Parity
+codec = { package = "parity-scale-codec", version = "1.3", default-features = false, features = ["derive", "full"] }
+kvdb = "0.7"
+kvdb-rocksdb = "0.9"
+hash-db = "0.15"
+parity-util-mem = "0.7"
+
+
+# Substrate
+sp-database = { git = "https://github.com/paritytech/substrate", package="sp-database" , rev = "647ad15565d7c35ecf00b73b12cccad9858780b9" }
```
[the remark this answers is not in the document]
insipx: I think all the substrate dependencies will be pointed at this rev (it's the latest substrate that polkadot is using)

I was actually just in the process of updating everything to polkadot v0.8.25 :P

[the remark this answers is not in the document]
insipx: Also on rustfmt -- I think the .rustfmt.toml has been committed for a while already -- it's just a matter of setting up our editors to use it and also have it in the CI pipeline to fail if code isn't formatted correctly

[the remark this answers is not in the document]
dvdplm: re-rustfmt: lovely, for 11 months! But without failing CI it's perhaps not as useful as it could be... ;)
Also, what do you think about having a common "tools team" rustfmt? Perhaps the same as `jsonrpsee` (which comes from `parity-common`):

```toml
hard_tabs = true
max_width = 120
use_small_heuristics = "Max"
edition = "2018"
```

As for `rev`, I still think we need a comment documenting why it's set to what it's set to.

insipx:
Haven't given this a full review yet, but I think I can answer this one:
>     * [ ]  Should `/bin/` crates be included in workspace?

I'm of the opinion that /bin/polkadot-archive and node-template-archive should live outside of the workspace. It should make it easier to manage dependencies per-project

substrate-archive-backend/Cargo.toml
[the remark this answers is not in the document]
dvdplm: It's a bit irksome that all of `archive` has to depend on specific substrate versions. It would be good if at least some of these new crates could be agnostic to the substrate version, so that only the binary crates would need to be built against a specific version. I don't think that's possible right now (or ever), but this tight coupling is a bit of a code smell.

substrate-archive-backend/src/lib.rs
insipx: If someone wants to use their own logger other than `fern` they could instantiate it themselves instead of using the archive default

substrate-archive/Cargo.toml
niklasad1: that's not introduced in this PR but `flume = "=0.7.1` seems like a greate idea without having any clue about the context here.

/cc @insipx I suppose you can explain this :)

insipx: yeah, flume =0.7.1 should be the way to go

When flume first updated with 0.7.2 (despite the minor release), there was a breaking change in the function of their asynchronous `send_async` and `recv_async` methods because of the use of a `cell` in one of the Sinks. This resulted in not being able to use recv_async across futures bounds. Flume is on 0.9 now, however, and I believe that the issue has been resolved. 

This update should just be done in another PR though, it touches basically all of the actors and some more

substrate-archive-backend/src/database.rs
```diff
@@ -56,7 +57,16 @@ impl ReadOnlyDatabase {
         })
     }
 
-    pub fn get(&self, col: ColumnId, key: &[u8]) -> Option<Vec<u8>> {
+    pub fn catch_up_count(&self) -> Option<usize> {
+        if !self.track_catchups {
+            log::warn!("catchup tracking is not enabled");
+            None
+        } else {
+            Some(self.catch_counter.fetch_add(0, Ordering::Relaxed))
```
niklasad1, line 65: not really introduced in this PR but `self.catch_counter.fetch_add(0, Ordering::Relaxed)` looks strange, it's better to use `load` if we just want read the value.

substrate-archive-common/src/util.rs
niklasad1: The "error message" could also be improved, https://docs.rs/directories/3.0.1/directories/struct.BaseDirs.html#method.new fails if the home directory couldn't be found.

substrate-archive/src/types.rs
```diff
@@ -0,0 +1,47 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with substrate-archive.  If not, see <http://www.gnu.org/licenses/>.
+
+use sp_runtime::traits::Block as BlockT;
+use substrate_archive_common::error::Result;
+
+pub trait ThreadPool: Send + Sync {
```
niklasad1, line 20: would perhaps be better to rename this file to `traits` and perhaps move `ReadOnlyDatabaseTrait` to this file.

substrate-archive/src/types.rs
```diff
@@ -0,0 +1,47 @@
...
+    type In: Send + Sync + std::fmt::Debug;
+    type Out: Send + Sync + std::fmt::Debug;
+    /// Adds a task to the threadpool.
+    /// Should not block!
+    fn add_task(&self, d: Vec<Self::In>, tx: flume::Sender<Self::Out>) -> Result<usize>;
+}
+
+#[async_trait::async_trait(?Send)]
+pub trait Archive<B: BlockT + Unpin>
+where
```
niklasad1, line 30: Would make sense to remove the `generic trait param`  on the entire trait and just add the bounds on context because it's not needed by the other methods:

```rust
trait Archive {
    // other methods...

    fn context<B>(&self) -> Result<super::actors::ActorContext<B>>
    where 
           B: BlockT + Unpin,
           <B as BlockT>::Hash: Unpin;
```

FWIW, it would likely require some additional refactoring and probably not worth the effort

substrate-archive-backend/src/util.rs
```diff
@@ -118,7 +117,7 @@ pub mod meta_keys {
 }
 
 pub fn read_header<Block: BlockT>(
-    db: &ReadOnlyDatabase,
+    db: Arc<dyn ReadOnlyDatabaseTrait>,
```
niklasad1, line 120: AFAIU the introduced `Arc<dyn ReadOnlyDatabaseTrait>` seems needless, you should be able to have `&dyn ReadOnlyDatabaseTrait` and use `Arc::deref ` i.e, `&*self.db`

substrate-archive-backend/src/read_only_backend/blockchain_backend.rs
```diff
@@ -94,13 +99,13 @@ impl<Block: BlockT> BlockchainBackend<Block> for ReadOnlyBackend<Block> {
...
     fn header(&self, id: BlockId<Block>) -> ChainResult<Option<Block::Header>> {
-        util::read_header::<Block>(&self.db, columns::KEY_LOOKUP, columns::HEADER, id)
+        util::read_header::<Block>(self.db.clone(), columns::KEY_LOOKUP, columns::HEADER, id)
             .map_err(|e| BlockchainError::Msg(e.to_string()))
     }
 
     fn info(&self) -> Info<Block> {
         // TODO: Remove expect
-        let meta = util::read_meta::<Block>(&self.db, columns::HEADER)
+        let meta = util::read_meta::<Block>(self.db.clone(), columns::HEADER)
```
niklasad1, line 108: if you agree with changing: 
```rust
pub fn read_meta<Block>(
     db: &dyn ReadOnlyDatabaseTrait,
     col_header: u32,
 ) 
```

```suggestion
        let meta = util::read_meta::<Block>(&*self.db, columns::HEADER)
```

substrate-archive/src/types.rs
insipx, line 20: Not from this PR, but we can just get rid of the `ThreadPool` trait, it's not being used for anything anymore

substrate-archive-backend/src/database.rs
```diff
@@ -75,27 +85,24 @@ impl ReadOnlyDatabase {
             }
         }
     }
+}
 
-    pub fn iter<'a>(&'a self, col: u32) -> impl Iterator<Item = KeyValuePair> + 'a {
-        self.inner.iter(col)
+impl ReadOnlyDatabaseTrait for ReadOnlyDatabase {
```
insipx, line 90: Now that we have a `ReadOnlyDatabaseTrait`, we should rename the `ReadOnlyDatabase` to something else. Perhaps `ReadOnlyRocksDB` or `RockDBSecondary` or something (not sure if i'm 100% happy with either of those) and then rename `ReadOnlyDatabaseTrait` to `ReadOnlyDatabase`. Should make it clearer on what's actually happening

substrate-archive-common/src/util.rs
insipx: It's this method: https://docs.rs/directories/3.0.1/directories/struct.BaseDirs.html#method.data_local_dir

so 

Platform | Value | Example
-- | -- | --
Linux | $XDG_DATA_HOME or $HOME/.local/share | /home/alice/.local/share
macOS | $HOME/Library/Application Support | /Users/Alice/Library/Application Support
Windows | {FOLDERID_LocalAppData} | C:\Users\Alice\AppData\Local


and I think it's only used for the log directory so far, however it is a hard-coded path so we should probably update the comment with that

insipx, line 108:
```suggestion
/// Get the path to a local substrate directory where we can save data.
/// Platform | Value | Example
/// -- | -- | --
/// Linux | $XDG_DATA_HOME or $HOME/.local/share/substrate_archive | /home/alice/.local/share/substrate_archive/
/// macOS | $HOME/Library/Application Support/substrate_archive | /Users/Alice/Library/Application Support/substrate_archive/
/// Windows | {FOLDERID_LocalAppData}\substrate_archive | C:\Users\Alice\AppData\Local\substrate_archive
```

substrate-archive-backend/src/read_only_backend.rs
```diff
@@ -44,23 +42,25 @@ use sp_runtime::{
     Justification,
 };
 use std::{convert::TryInto, sync::Arc};
+use substrate_archive_common::database::ReadOnlyDatabaseTrait;
+use substrate_archive_common::error::Result;
 
 pub struct ReadOnlyBackend<Block: BlockT> {
-    db: Arc<ReadOnlyDatabase>,
+    db: Arc<dyn ReadOnlyDatabaseTrait>,
```
maciejhirsz, line 49: Just to state the obvious here: are we concerned about the binary size, or is there another reason to use dyn dispatch instead of static generics? Is the database access so infrequent we don't care about jumping through vtables for it?

substrate-archive-common/src/database.rs
```diff
@@ -0,0 +1,13 @@
+use super::error::Result;
+
+pub type KeyValuePair = (Box<[u8]>, Box<[u8]>);
+
+// Archive specific K/V database reader implementation
+pub trait ReadOnlyDatabaseTrait: Send + Sync {
```
maciejhirsz, line 6: There is a minor naming inconsistency with imported substrate code here which uses `NameT` rather than `NameTrait` for traits.

maciejhirsz (commented):
One question, and one minor nitpick

emostov:
The most recent commit is not viable to merge as I am in the middle of trying to use https://github.com/paritytech/substrate-archive/blob/7626deb8713aa4ffdadac55054ac0044ac6c50cd/substrate-archive-common/src/database.rs#L32

in response to @maciejhirsz feedback regarding using dynamic dispatch. I am not sure if this is idiomatic (or if it will even work) but will try and and keep working on it tomorrow since it seems like a worthwhile refactor. I am thinking this might make sense in another PR since it touches so many parts, but not sure if worth the extra effort.

I also am planning on removing the changes (silence clippy warnings) to `substrate-archive/src/database/querys.rs` and putting them in a separate PR since they are completely unrelated.

Not sure why the CI is failing on `polkadot-service`. I have been having that fail a lot locally as well, but not every time.

substrate-archive/src/actors.rs
```diff
@@ -343,18 +342,12 @@ where
     fn shutdown(self) -> Result<()> {
         let _ = self.kill_tx.send(());
         self.handle.join()?;
-        if let Some(c) = self.context.backend().backing_db().catch_up_count() {
```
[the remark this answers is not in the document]
insipx: Yeah, if we remove catch-up count here we should also remove the logic in the rocksdb-specific database implementation. I think it's OK to remove, was mostly for debugging purposes anyway

substrate-archive-backend/src/read_only_backend/state_backend.rs
```diff
@@ -24,7 +24,7 @@ use sp_runtime::traits::{Block as BlockT, HashFor};
 use sp_state_machine::{StateMachineStats, TrieBackend, UsageInfo as StateUsageInfo};
 use std::marker::PhantomData;
 use std::sync::Arc;
-use substrate_archive_common::database::{ReadOnlyDB, ReadOnlyDBContainer};
+use substrate_archive_common::database::{ReadOnlyDB};
```
niklasad1, line 27:
```suggestion
use substrate_archive_common::database::ReadOnlyDB;
```

substrate-archive-backend/src/database.rs
```diff
@@ -136,7 +143,7 @@ impl<H: Clone> DatabaseTrait<H> for ReadOnlyDatabase {
 }
 
 /*
-impl KeyValueDB for ReadOnlyDatabase {
+impl KeyValueDB for ReadOnlyDB {
```
niklasad1, line 146: There's some "uncommented code` here, why? Can it be removed?

substrate-archive-backend/src/frontend.rs
```diff
@@ -17,44 +17,44 @@
...
-// use sc_client_db::Backend;
-use self::executor::ArchiveExecutor;
-use crate::{backend::database::ReadOnlyDatabase, error::Error as ArchiveError};
-use futures::{task::SpawnExt, Future};
 use sc_executor::{NativeExecutionDispatch, NativeExecutor, WasmExecutionMethod};
 use sp_api::ConstructRuntimeApi;
 use sp_core::traits::SpawnNamed;
 use sp_runtime::traits::{BlakeTwo256, Block as BlockT};
 use std::sync::Arc;
+use substrate_archive_common::database::{ReadOnlyDB};
```
niklasad1, line 31:
```suggestion
use substrate_archive_common::database::ReadOnlyDB;
```

substrate-archive-common/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "substrate-archive-common"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+
+[dependencies]
+rayon = "1.4"
```
niklasad1, line 8: you need to add the `bincode` dependency here in order for it compile :)

substrate-archive/src/actors.rs
```diff
@@ -42,30 +39,32 @@ use sp_runtime::traits::{Block as BlockT, Header as _, NumberFor};
...
+pub use substrate_archive_common::{msg, msg::Die, Result, database::ReadOnlyDB};
 use xtra::prelude::*;
 
 // TODO: Split this up into two objects
 // System should be a factory that produces objects that should be spawned
 
 /// Context that every actor may use
 #[derive(Clone)]
-pub struct ActorContext<B: BlockT + Unpin>
+pub struct ActorContext<B: BlockT + Unpin, D: ReadOnlyDB + 'static>
```
niklasad1, line 51: it looks like you need to add `Clone` bound in order for this to work after changing to `static dispatch` because it is `cloned` in the Archive impl.

substrate-archive/src/actors.rs
```diff
@@ -92,44 +91,46 @@ where
...
     metadata: Address<workers::Metadata<B>>,
     db_pool: Address<ActorPool<DatabaseActor<B>>>,
 }
 
 /// Control the execution of the indexing engine.
 /// Will exit on Drop.
-pub struct System<B, R, C>
+pub struct System<B, R, C, D>
 where
+    D: ReadOnlyDB + 'static,
```
niklasad1, line 109: you can remove all the trait bounds on the struct and just add them on each impl.

substrate-archive/src/types.rs
niklasad1: cool, sounds good to me.

substrate-archive-backend/src/database.rs
insipx: It can be removed, something I left in the code from a while back

substrate-archive-backend/src/util.rs
```diff
@@ -70,8 +68,8 @@ pub fn open_database(
         NUM_COLUMNS,
         other_col_budget,
     );
-    super::database::ReadOnlyDatabase::open(db_config, &path)
-        .map_err(|err| sp_blockchain::Error::Backend(format!("{:?}", err)))
+    super::database::SecondaryRocksDB::open(db_config, &path)
+        .map_err(|err| sp_blockchain::Error::Backend(format!("{:?}", err))) as
```
insipx, line 72:
```suggestion
        .map_err(|err| sp_blockchain::Error::Backend(format!("{:?}", err)))
```

substrate-archive-backend/src/read_only_backend.rs
insipx: TBH it didn't cross my mind to use static generics when designing the ReadOnlyBackend the first time around, and it might be a worthy endeavour so I created #115. Binary size shouldn't be much of an issue since we aren't running on very constrained devices but rather servers that should have more than enough storage space.

insipx:
> The most recent commit is not viable to merge as I am in the middle of trying to use
> 
> https://github.com/paritytech/substrate-archive/blob/7626deb8713aa4ffdadac55054ac0044ac6c50cd/substrate-archive-common/src/database.rs#L32
> 
> in response to @maciejhirsz feedback regarding using dynamic dispatch. I am not sure if this is idiomatic (or if it will even work) but will try and and keep working on it (need to add the type and bounds everywhere) tomorrow since it seems like a worthwhile refactor . I am thinking this might make sense in another PR since it touches so many parts, but not sure if worth the extra effort.
> 


I think a separate PR for dispatch would be good, since this PR is already fairly large, I made #115  for it

> Not sure why the CI is failing on `polkadot-service`. I have been having that fail a lot locally as well, but not every time.

if you merge in #113 that should fix the CI I think

niklasad1:
> I think a separate PR for dispatch would be good, since this PR is already fairly large, I made #115 for it

I agree but I think it might be too late since the implementation for static dispatch is almost finished AFAIK.

insipx:
> > I think a separate PR for dispatch would be good, since this PR is already fairly large, I made #115 for it
> 
> I agree but I think it might be too late since the implementation for static dispatch is almost finished AFAIU.

Ah you are right, in that case it doesn't matter too much then :)

substrate-archive/src/actors.rs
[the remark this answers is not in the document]
emostov: Do we want to remove `catch_counter` and `track_catchups` from the `SecondaryRocksDB` struct as well? If so I might make an issue and just do it in another pr.

substrate-archive/src/database/queries.rs
```diff
@@ -64,6 +64,7 @@ pub(crate) async fn missing_blocks_min_max(
 ) -> Result<HashSet<u32>> {
     let min = i32::try_from(min).unwrap_or(i32::MAX);
     let max_block_load = i64::try_from(max_block_load).unwrap_or(i64::MAX);
+    #[allow(clippy::toplevel_ref_arg)]
```
niklasad1, line 67: this is caused by `sqlx::query_as!` and not by us, correct?

It just a [style warning](https://rust-lang.github.io/rust-clippy/master/index.html) but perhaps open an issue upstream?!

substrate-archive/src/archive.rs
```diff
@@ -33,14 +27,17 @@ use sp_runtime::{
     traits::{BlakeTwo256, Block as BlockT, NumberFor},
 };
 use std::{marker::PhantomData, path::PathBuf, sync::Arc};
+// use substrate_archive_backend::{open_database, runtime_api, ReadOnlyBackend, TArchiveClient, SecondaryRocksDB};
```
niklasad1, line 30:
```suggestion
```

substrate-archive-backend/src/read_only_backend.rs
emostov: Done, transitioned to static generics.

substrate-archive/src/database/queries.rs
emostov: There already is an issue up for it https://github.com/launchbadge/sqlx/issues/594

emostov:
I think this is ready for another review now

substrate-archive-backend/src/frontend.rs
```diff
@@ -17,19 +17,19 @@
...
-use self::executor::ArchiveExecutor;
-use crate::{backend::database::ReadOnlyDatabase, error::Error as ArchiveError};
-use futures::{task::SpawnExt, Future};
 use sc_executor::{NativeExecutionDispatch, NativeExecutor, WasmExecutionMethod};
 use sp_api::ConstructRuntimeApi;
 use sp_core::traits::SpawnNamed;
 use sp_runtime::traits::{BlakeTwo256, Block as BlockT};
 use std::sync::Arc;
+use substrate_archive_common::database::ReadOnlyDatabaseTrait;
+use substrate_archive_common::error::Error as ArchiveError;
```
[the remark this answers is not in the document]
emostov: https://github.com/paritytech/substrate-archive/issues/116

substrate-archive-common/src/database.rs
niklasad1: Right, we could refactor this later with an associate type `Error` for defining implementation related errors.

niklasad1 (approved):
LGTM, nice work

substrate-archive-common/src/database.rs
```diff
@@ -0,0 +1,40 @@
...
+pub type KeyValuePair = (Box<[u8]>, Box<[u8]>);
+
+pub const NUM_COLUMNS: u32 = 11;
+
+// Archive specific K/V database reader implementation
+pub trait ReadOnlyDB: Send + Sync {
+    /// Read key/value pairs from the database
+    fn get(&self, col: u32, key: &[u8]) -> Option<Vec<u8>>;
+    /// Iterate over all blocks in the database
+    fn iter<'a>(&'a self, col: u32) -> Box<dyn Iterator<Item = KeyValuePair> + 'a>;
```
maciejhirsz, line 29: We could get rid of the Box<dyn ...> here as well:

```rust
pub trait ReadOnlyDB: Send + Sync {
    /// Iterator type to be returned
    type Iter: Iterator<Item = KeyValuePair>;
    /// Read key/value pairs from the database
    fn get(&self, col: u32, key: &[u8]) -> Option<Vec<u8>>;
    /// Iterate over all blocks in the database
    fn iter(&self, col: u32) -> Self::Iter;
```

Unfortunately, `kvdb_rocksdb` uses `impl Iterator` as a return type which obfuscates the concrete iterator type it's returning, which breaks defining the associated type on the trait impl. This would be a small thing to fix as a separate PR later on if the signature is fixed upstream.

maciejhirsz (approved):
:ok_hand: 

There is still one dynamic dispatch I'd get rid of, but that requires upstream changes and I don't think is worth blocking this PR atm.

substrate-archive-backend/src/database.rs
```diff
@@ -75,33 +77,68 @@ impl ReadOnlyDatabase {
...
-            None
-        } else {
-            Some(self.catch_counter.fetch_add(0, Ordering::Relaxed))
+    fn open_database(
+        path: &str,
+        cache_size: usize,
+        db_path: PathBuf,
+    ) -> sp_blockchain::Result<SecondaryRocksDB> {
+        // need to make sure this is `Some` to open secondary instance
+        let db_path = db_path.as_path().to_str().expect("Creating db path failed");
```
dvdplm, line 105: Can this be something like?

```suggestion
        let db_path = db_path.as_path().to_str().map_err(|e| Error::Msg(format!("Bad database path: {:?}", e)?;
```

substrate-archive-backend/src/database.rs
```diff
@@ -17,13 +17,15 @@
...
 use std::sync::atomic::{AtomicUsize, Ordering};
-
-pub type KeyValuePair = (Box<[u8]>, Box<[u8]>);
+use substrate_archive_common::{
+    database::{KeyValuePair, ReadOnlyDB, NUM_COLUMNS},
+    Result,
+};
 
 pub struct Config {
     /// track how many times try_catch_up_with_primary is called
```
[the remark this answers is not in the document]
dvdplm: We should document what "catching up" means in archive. Or maybe it's doc'd already?

substrate-archive-backend/src/database.rs
```diff
@@ -75,33 +77,68 @@ impl ReadOnlyDatabase {
...
+        // need to make sure this is `Some` to open secondary instance
+        let db_path = db_path.as_path().to_str().expect("Creating db path failed");
+        let mut db_config = Config {
+            track_catchups: true,
+            config: DatabaseConfig {
+                secondary: Some(db_path.to_string()),
+                ..DatabaseConfig::with_columns(NUM_COLUMNS)
+            },
+        };
+        let state_col_budget = (cache_size as f64 * 0.9) as usize;
```
dvdplm, line 113: I wonder if this makes sense for this use case. Like I have literally no idea what the best values are here.

substrate-archive-backend/src/database.rs
```diff
@@ -75,33 +77,68 @@ impl ReadOnlyDatabase {
...
+        Self::open(db_config, &path)
+            .map_err(|err| sp_blockchain::Error::Backend(format!("{:?}", err)))
     }
 }
 
 type DBError = std::result::Result<(), sp_database::error::DatabaseError>;
 //TODO: Remove panics with a warning that database has not been written to / is read-only
-/// Preliminary trait for ReadOnlyDatabase
-impl<H: Clone> DatabaseTrait<H> for ReadOnlyDatabase {
+/// Preliminary trait for ReadOnlyDB
```
dvdplm, line 140: This doc string needs to be rewritten.

substrate-archive-backend/src/frontend.rs
```diff
@@ -17,44 +17,43 @@
...
-// use sc_client_db::Backend;
-use self::executor::ArchiveExecutor;
-use crate::{backend::database::ReadOnlyDatabase, error::Error as ArchiveError};
-use futures::{task::SpawnExt, Future};
 use sc_executor::{NativeExecutionDispatch, NativeExecutor, WasmExecutionMethod};
 use sp_api::ConstructRuntimeApi;
 use sp_core::traits::SpawnNamed;
 use sp_runtime::traits::{BlakeTwo256, Block as BlockT};
 use std::sync::Arc;
+use substrate_archive_common::{database::ReadOnlyDB, error::Error as ArchiveError};
```
dvdplm, line 31: Re-export `database::ReadonlyDB` and `error::Error`? so this can be `use substrate_archive_common::{ReadOnlyDB, Error as ArchiveError}`?

substrate-archive-common/src/util.rs
```diff
@@ -10,50 +10,17 @@
...
-        path.push("substrate_archive");
-        path
-    } else {
-        panic!("Couldn't establish substrate data local path");
-    }
-}
+use std::path::{Path, PathBuf};
 
 #[cfg(feature = "logging")]
 pub fn init_logger(std: log::LevelFilter, file: log::LevelFilter) {
```
dvdplm, line 26: Document panicking behaviour please.

substrate-archive-common/src/util.rs
```diff
@@ -64,10 +31,13 @@ pub fn init_logger(std: log::LevelFilter, file: log::LevelFilter) {
...
         .trace(Color::Magenta);
 
-    // let mut log_dir = dirs::data_local_dir().expect("failed to find local data dir for logs");
-    // log_dir.push("substrate_archive");
-    let mut log_dir = substrate_dir();
-    create_dir(log_dir.as_path());
+    let mut log_dir = match substrate_dir() {
+        Err(e) => panic!(e),
+        Ok(dir) => dir,
+    };
```
dvdplm, line 37: This is the same as `let log_dir = substrate_dir().unwrap()` right?

substrate-archive-common/src/util.rs
```diff
@@ -64,10 +31,13 @@ pub fn init_logger(std: log::LevelFilter, file: log::LevelFilter) {
...
+    if create_dir(log_dir.as_path()).is_err() {
+        panic!("A directory in the path of the substrate directory could not be created")
+    }
```
dvdplm, line 40: I see what you're doing here, and I think it's better than before. Ideally I think we should hand the error ever upwards, i.e. make `init_logger()` return `Result` and so forth all the way up to the main binary. That's where we want the panics to happen.

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +108,37 @@ fn format_opt(file: Option<String>) -> String {
...
+/// macOS | $HOME/Library/Application Support/substrate_archive | /Users/Alice/Library/Application Support/substrate_archive/
+/// Windows | {FOLDERID_LocalAppData}\substrate_archive | C:\Users\Alice\AppData\Local\substrate_archive
+pub fn substrate_dir() -> Result<PathBuf, &'static str> {
+    if let Some(base_dirs) = dirs::BaseDirs::new() {
+        let mut path = base_dirs.data_local_dir().to_path_buf();
+        path.push("substrate_archive");
+        Ok(path)
+    } else {
+        Err("No valid home directory path could be retrieved from the operating system")
+    }
```
dvdplm, line 125: Use `match` and use the actual error from `BaseDirs::new()`?

substrate-archive-common/src/util.rs
```diff
@@ -138,3 +108,37 @@ fn format_opt(file: Option<String>) -> String {
...
+}
+
+/// Create an arbitrary directory on disk.
+pub fn create_dir(path: &Path) -> Result<(), &'static str> {
+    if let Err(e) = std::fs::create_dir_all(path) {
+        match e.kind() {
+            std::io::ErrorKind::AlreadyExists => (),
+            _ => return Err("A directory in path could not be created"),
```
dvdplm, line 133: It would be better to capture the error and return it, worst case as `Err(format!("oh noes, badness: {:?}", e))`.

substrate-archive/Cargo.toml
```diff
@@ -0,0 +1,83 @@
...
+tempfile = "3.1.0"
+async-executor = "0.1.2"
+arc-swap = "0.4.7"
+
+# Workspace
+substrate-archive-backend = { path = '../substrate-archive-backend' }
+substrate-archive-common = { path = '../substrate-archive-common' }
+
+[dev-dependencies]
+# desub-extras = { path = "../desub/extras/", package="desub-extras", features = ["polkadot"] }
```
dvdplm, line 70:
```suggestion
```

substrate-archive/src/database/queries.rs
dvdplm, line 67:
```suggestion
    // Remove after launchbadge/sqlx#594 is fixed
    #[allow(clippy::toplevel_ref_arg)]
```

dvdplm (commented):
Real close now. Left some whines.

I guess the biggest issue I have is it doesn't compile on latest nightly.

Cargo.toml
```diff
@@ -1,83 +1,10 @@
...
+[workspace]
+members = [
+  "substrate-archive",
+  "substrate-archive-backend",
+  "substrate-archive-common"
+]
+exclude = [
+  "bin/polkadot-archive",
+  "bin/node-template-archive"
+]
```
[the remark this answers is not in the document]
emostov: I actually could not compile without specifically including and excluding all the crates within nested in root.

substrate-archive-common/src/util.rs
emostov: #122

substrate-archive-common/src/util.rs
emostov: #122

substrate-archive-common/src/util.rs
emostov: #122

substrate-archive-backend/src/database.rs
emostov: I have no idea. @insipx ?

insipx: Been a while since I played with these values, but as far as I could tell, increasing the cache size or giving a ridiculously large cache size didn't really increase performance other than just use more memory. Inversely, giving no cache size breaks the rocksdb portions of archive.

So, in the end I went with the numbers as they were in substrate (I think this multiplication is copied right over). IIRC state column budget describes how much cache is given to each column, but we're not really accessing the same value multiple times in archive. The only time that happens is with the blocks actor/blocks column, where we might potentially iterate over the same block number in the `key` column looking for a block that is missing.

substrate-archive-backend/src/database.rs
emostov: #122

substrate-archive-common/src/database.rs
emostov: #123

substrate-archive/src/actors.rs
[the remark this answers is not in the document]
emostov: #124 - feel free to close issue if it doesn't make sense

substrate-archive-backend/src/database.rs
```diff
@@ -32,20 +34,20 @@ pub struct Config {
 }
 
 #[derive(parity_util_mem::MallocSizeOf)]
-pub struct ReadOnlyDatabase {
+pub struct SecondaryRocksDB {
     inner: Database,
     catch_counter: AtomicUsize,
```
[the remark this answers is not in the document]
emostov: #124

emostov:
Made several issues for most of the final comments. Maybe @insipx or @dvdplm wants to do final review and merge? Note: I likely won't be able to work on this after Oct 30

insipx (approved):
LGTM! good job. I'll merge it in

PR #118, opened 2020-10-29
Use substrate master

dvdplm:
I don't know if this is a viable solution but it compiles. :)

If at all possible I think it is worthwhile to stick as closely to upstream as possible. That is what polkadot does I believe.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,23 +43,23 @@ kvdb-rocksdb = "0.9"
 codec = { package = "parity-scale-codec", version = "1.3", default-features = false, features = ["derive", "full"] }
 hash-db = "0.15"
 
-# Substrate is pinned to the latest substrate version that polkadot release v0.8.25 is
+# Substrate is pinned to the latest substrate version that polkadot release v0.8.26 is
 # using. This is necessary to make substrate-archive compatible with the polkadot runtime.
-sp-database = { git = "https://github.com/paritytech/substrate", package = "sp-database", rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-storage = { git = "https://github.com/paritytech/substrate", package = "sp-storage" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-externalities = { git = "https://github.com/paritytech/substrate", package = "sp-externalities" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -43,23 +43,23 @@ kvdb-rocksdb = "0.9"
...
-sp-block-builder = { git = "https://github.com/paritytech/substrate", package = "sp-block-builder" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-version = {  git = "https://github.com/paritytech/substrate", package = "sp-version" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sc-executor = { git = "https://github.com/paritytech/substrate", package = "sc-executor" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sc-chain-spec = { git = "https://github.com/paritytech/substrate", package = "sc-chain-spec" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-trie = { git = "https://github.com/paritytech/substrate", package = "sp-trie" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-state-machine = { git = "https://github.com/paritytech/substrate", package = "sp-state-machine" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
-sp-io = { git = "https://github.com/paritytech/substrate", package = "sp-io" , rev = "7314a78d49d0dc3862c6ef7c8115cb1b3d0c0fd2" }
+sp-database = { git = "https://github.com/paritytech/substrate", package = "sp-database", branch = "master" }
```
liuchengxu, line 48: Unrelated to this PR, but isn't `package = "sp-database"` a bit of redundant since we are not using any alias here? Just saying :-

insipx, line 48:
```suggestion
sp-database = { git = "https://github.com/paritytech/substrate", branch = "master" }
```

insipx:
We used to track substrate and polkadot master before 0.8.24 I think, so for the most part I don't have anything against tracking master again

Every now and  then there's just some build issues from one or the other repository that might be tricky to resolve, but it seems like it might still result in less maintenance then pinning the versions

insipx (approved):

niklasad1 (approved):

Cargo.toml
dvdplm: You're perfectly right. TBH I had to stare on the code for quite a bit before I understood what you meant, it's like selective blindness. Fixed. :)

PR #126, opened 2020-11-02
Refactor errors from `create_dir`, `init_logger`, `substrate_dir`

### LINKED ISSUE #122: Refactor `create_dir`, `substrate_dir`, and `init_logger` to return `Result` to main binary
Ideally I think we should hand the error ever upwards, i.e. make `init_logger()` return `Result` and so forth all the way up to the main binary. That's where we want the panics to happen.

Also, where possible, it would be better to catch the actual errors returned instead of adding our own.

Another place we can pass an error up to main binary is here: https://github.com/paritytech/substrate-archive/pull/108#discussion_r514205573

_Originally posted by @dvdplm in https://github.com/paritytech/substrate-archive/pull/108#r514212663_

insipx:
This PR also refactors `create_database_path` to use the `substrate_dir` function

fixes #122


substrate-archive/src/archive.rs
```diff
@@ -161,18 +161,11 @@ fn parse_urls(chain_data_path: Option<String>, pg_url: Option<String>) -> (Strin
...
-            path
-        } else {
-            panic!("Couldn't establish substrate data local path");
-        };
-        std::fs::create_dir_all(path.as_path())
-            .expect("Unable to create rocksdb secondary directory");
+        let mut path = substrate_archive_common::util::substrate_dir()?;
+        path.push("rocksdb_secondary");
+        path.push(chain);
+        path.push(id);
```
dvdplm, line 167:
```suggestion
        let mut path = util::substrate_dir()?.extend(&["rocksdb_secondary", chain, id]);
```

maybe?

substrate-archive-common/src/util.rs
```diff
@@ -112,22 +98,28 @@ fn format_opt(file: Option<String>) -> String {
...
 /// Create an arbitrary directory on disk.
-pub fn create_dir(path: &Path) -> Result<(), &'static str> {
+pub fn create_dir(path: &Path) -> Result<(), ArchiveError> {
     if let Err(e) = std::fs::create_dir_all(path) {
         match e.kind() {
             std::io::ErrorKind::AlreadyExists => (),
-            _ => return Err("A directory in path could not be created"),
+            _ => {
+                return Err(ArchiveError::from(
+                    "A directory in path could not be created",
```
dvdplm, line 120:
```suggestion
                    format!("A directory in the path '{:?}' could not be created", path),
```

maybe?

dvdplm (approved):
tight.

niklasad1 (approved):

PR #127, opened 2020-11-02
Remove Catchup tracking

### LINKED ISSUE #124: Remove `catch_counter` and `track_catchups` from `SecondaryRocksDB`
Keeping track of catch ups by the secondary RocksDB was primarily for debugging and is no longer in use. The relevant functionality for tracking catch ups can be removed from the `SecondaryRocksDB`.

_Originally posted by @emostov in https://github.com/paritytech/substrate-archive/pull/108#discussion_r513705070_

insipx:
Removes tracking database 'catch-ups' in RocksDB-specific code.

fixes #124


substrate-archive-backend/src/database.rs
```diff
@@ -104,7 +92,6 @@ impl ReadOnlyDB for SecondaryRocksDB {
         // need to make sure this is `Some` to open secondary instance
         let db_path = db_path.as_path().to_str().expect("Creating db path failed");
         let mut db_config = Config {
-            track_catchups: true,
```
dvdplm, line 107: Is this a breaking change? Like, did we expose this option? If yes, it's changelog-worthy.

dvdplm (approved):
:ketchup: is no more.

niklasad1 (approved):

substrate-archive-backend/src/database.rs
niklasad1: Technically, no because `Config` is not re-exported but I think it should be re-exported because otherwise, it not possible to instantiate `SecondaryRocksDB` AFAIU.

insipx: Yeah it was never directly exposed, only printed on shutdown

PR #129, opened 2020-11-06
Update Cargo.lock for polkadot 0.8.26

insipx:
Also fixes an import that caused polkadot-archive to fail to build

also seems like i missed running rustfmt on the `polkadot-archive` submodule/project so I included that in this PR too

### DIFF
--- a/bin/polkadot-archive/Cargo.toml
+++ b/bin/polkadot-archive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "polkadot-archive"
-version = "0.2.1"
+version = "0.2.2"
 authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
 edition = "2018"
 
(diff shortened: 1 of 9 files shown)


niklasad1 (approved):

bin/polkadot-archive/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "polkadot-archive"
-version = "0.2.1"
+version = "0.2.2"
 authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
 edition = "2018"
```
emostov, line 6: It looks like this is just on the latest version of polkadot maser - what pins it to 0.8.26?

emostov (approved):

bin/polkadot-archive/Cargo.toml
insipx: Yeah, since we're just tracking polkadot master now we rely on the Cargo.lock for the correct version:

https://github.com/paritytech/substrate-archive/pull/129/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eR2667

PR #133, opened 2020-11-16
Use substrate LocalCallExecutor

insipx:
Since `LocalCallExecutor` is public now we can delete `executor.rs` in frontend. 

In addition, this release finally switches sqlx from beta to 0.4.0 (recently released).


dvdplm (approved):
NICE. I love it when we delete code. :)

emostov (approved):

PR #92, opened 2020-09-02
[WIP] State Tracing

insipx:
Preliminary state tracing support for native runtimes

related:
https://github.com/paritytech/substrate/pull/6916
https://github.com/paritytech/substrate/pull/5826
https://github.com/paritytech/substrate/pull/6672

Issue #43, opened 2020-05-20
Should eliminate RPC dependencies entirely


insipx:
Since using a custom substrate client, an RPC is no longer needed and adds unnecessary bloat/complexity.

Should update the backend trait to use `BlockchainEvents` and `RuntimeApi/CallApiAt` providers. (`BlockchainEvents` for getting new finalized blocks, and `RuntimeAPI/CallApiAt` for getting the metadata.)

RuntimeApi may be tricky because of the associated types it requires

insipx:
This won't be possible, it's actually in our best interest to use the RPC in order to get new finalized blocks, and storage changes from those blocks (which comes directly from block_changes_notification) letting us avoid re-executing the block in order to get storage changes and keep up with the main chain.

So closing for now.

PR #134, opened 2020-11-18
State Tracing

insipx:
Continuation of work from #92 

Not ready for merge yet because:
- [x] migrations for tracing tables need to be created
- [x] Traces need insertion into Postgres Database
- [x] Documentation
- [x] Testing on Kusama and Polkadot 
- [x] Filtering by Target/Level

In this PR
- `ActorContext` -> `SystemConfig`
- Instantiating our own `native_executor_instance` in order to have access and change around the HostFunctions available in the executing WASM environment
- add a config option for WASM targets, add a CLI argument for the folder to override WASM


bin/polkadot-archive/src/config.rs
```diff
@@ -80,6 +88,7 @@ impl Config {
 			block_workers: toml_conf.as_ref().map(|c| c.block_workers).flatten(),
 			wasm_pages: toml_conf.as_ref().map(|c| c.wasm_pages).flatten(),
 			max_block_load: toml_conf.as_ref().map(|c| c.max_block_load).flatten(),
+			wasm_tracing: tracing,
```
dvdplm, line 91: Rename `tracing` to `wasm_tracing` perhaps?

bin/polkadot-archive/src/cli_opts.rs
```diff
@@ -41,6 +42,14 @@ impl CliOpts {
 
 		let chain = matches.value_of("chain").unwrap_or("polkadot");
 
-		CliOpts { file: file.map(|f| PathBuf::from(f)), log_level, log_num, chain: chain.to_string() }
+		let wasm_overrides = matches.value_of("wasm_overrides_path");
+
+		CliOpts {
+			file: file.map(|f| PathBuf::from(f)),
```
dvdplm, line 48: Could do the `map` on line #41

bin/polkadot-archive/src/cli_opts.rs
```diff
@@ -41,6 +42,14 @@ impl CliOpts {
...
+			log_level,
+			log_num,
+			chain: chain.to_string(),
```
dvdplm, line 51: Call `to_string()` on line #43?

bin/polkadot-archive/src/cli_opts.rs
```diff
@@ -41,6 +42,14 @@ impl CliOpts {
...
+			wasm_overrides_path: wasm_overrides.map(|p| PathBuf::from(p)),
```
dvdplm, line 52: Could do this above and have a cleaner retval here.

bin/polkadot-archive/src/archive.rs
```diff
@@ -57,12 +78,15 @@ pub fn run_archive<D: ReadOnlyDB + 'static>(config: Config) -> Result<Box<dyn Ar
 				..ArchiveBuilder::default()
 			}
 			.chain_data_db(db_path)
-			.chain_spec(spec)
-			.build()?;
+			.chain_spec(spec);
+			if let Some(t) = tracing {
```
dvdplm, line 82: Maybe just `if let Some(t) = config.wasm_tracing() {`

bin/polkadot-archive/test_conf.toml
```diff
@@ -23,9 +23,9 @@ cache_size = 128
 # else the wasm executor will run out of memory. This also increases substrate-archives
```
dvdplm, line 23: "This also increases substrate-archives memory usage"?

bin/polkadot-archive/test_conf.toml
```diff
@@ -44,3 +44,8 @@ db_pass = "123"
 westend_db = "westend-archive"
 kusama_db = "kusama-db"
 polkadot_db = "polkadot-db"
+
+#wasm_tracing_targets = '''
+#assets,atomic-swap,aura,authority-discovery,authorship,babe,balances,collective,contracts,democracy,elections,elections-phragmen,evm,executive,generic-asset,grandpa,identity,im-online,indices,membership,metadata,multisig,nicks,offences,proxy,randomness-collective-flip,recovery,scheduler,scored-pool,session,society,staking,sudo,support,system,timestamp,transaction-payment,treasury,utility,vesting,wasm_tracing,wasm-tracing,frame,frame-executive,frame_executive'''
+wasm_tracing_targets = '''wasm_tracing,wasm-tracing,support,executive,frame-executive,frame_executive,frame_support,frame-support,pallet_scheduler,pallet_society'''
```
dvdplm, line 50: `wasm_tracing` twice here

bin/node-template-archive/src/cli_opts.rs
```diff
@@ -19,48 +19,47 @@ use std::path::PathBuf;
...
+			0 => log::LevelFilter::Info,
+			1 => log::LevelFilter::Info,
+			2 => log::LevelFilter::Info,
+			3 => log::LevelFilter::Debug,
+			4 | _ => log::LevelFilter::Trace,
+		};
+		let file = matches.value_of("config").expect("Config is a required value");
+		let chain_spec;
+		let spec = matches.value_of("spec");
+		if spec.is_some() {
```
niklasad1, line 41: nitpick, pattern match directly spec and remove `if spec.is_some()`?

substrate-archive/src/archive.rs
```diff
@@ -52,6 +61,9 @@ pub struct Builder<B, R, D, DB> {
 	pub _marker: PhantomData<(B, R, D, DB)>,
 	/// maximimum amount of blocks to index at once
 	pub max_block_load: Option<u32>,
+	/// Enable state tracing while also specifying the targets
+	/// and directory where WASM is stored.
```
dvdplm, line 65: "…where the WASM runtimes are stored"?

substrate-archive-backend/src/block_exec.rs
```diff
@@ -106,7 +106,7 @@ where
 
 		let state = self.backend.state_at(self.id)?;
 
-		// FIXME: For some reason, wasm runtime calculates a different number of digest items
+		// Wasm runtime calculates a different number of digest items
 		// then what we have in the block
```
dvdplm, line 110: "than"

substrate-archive-common/src/util.rs
```diff
@@ -43,7 +43,7 @@ pub fn init_logger(std: log::LevelFilter, file: log::LevelFilter) -> Result<(),
 		.level_for("staking", log::LevelFilter::Warn)
 		.level_for("cranelift_codegen", log::LevelFilter::Warn)
 		.level_for("header", log::LevelFilter::Warn)
-		.level_for("", log::LevelFilter::Error)
+		// .level_for("", log::LevelFilter::Error)
```
dvdplm, line 46: We don't want errors logged?

insipx:
Didn't expect a review so fast! There are some problems with the current implementation I wrote up and was going to ask for feedback today

State tracing seems to work -- sort of. Replacing WASM code at runtime and tracing itself is working, but Substrate Archive is getting traces for only some parts of the substrate landscape: `sp_io::{hashing,allocator,crypto,storage,crypto}` but no top-level traces from e.g `frame_executive`/`frame_support`. I know that for some of these pallets, tracing hasn't been implemented in the runtime yet. However, I'd expect to get traces from at least executive for this line (for example): https://github.com/paritytech/substrate/blob/69198d1e5735798c8baa1921c6e2091c93260752/frame/executive/src/lib.rs#L295

Haven't made the migrations yet because i'm not sure if there might be something wrong with how i've setup the tracing. So far, it just logs the `sp_io::` targets outlined earlier as an INFO logline. For example this is a photo of two blocks executing and only getting trie/crypto tracing targets (with some filtered out because there were alot of loglines):

state tracing

substrate-archive-common/src/util.rs
insipx: Ah, this just filters out the target for fern. During my testing I was wondering if fern was maybe culling some of the state tracing logs, this line just tells fern to ignore any log with target "" and is above level Error

substrate-archive/src/actors/workers/state_tracing.rs
```diff
@@ -0,0 +1,106 @@
...
+impl Message for SpanMessage {
+	type Result = ();
+}
+
+#[async_trait::async_trait]
+impl Handler<SpanMessage> for TracingActor {
+	async fn handle(&mut self, msg: SpanMessage, _: &mut Context<Self>) {
+		let to_print = msg.0.target.as_str();
+		match to_print.as_ref() {
+			"sp_io::hashing" | "sp_io::allocator" | "sp_io::storage" => {}
```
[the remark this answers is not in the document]
insipx: Wrote it up in the latest comment

substrate-archive/src/actors/workers/state_tracing.rs
```diff
@@ -0,0 +1,106 @@
...
+struct EventMessage(TraceEvent);
+
+impl Message for EventMessage {
+	type Result = ();
+}
+
+#[async_trait::async_trait]
+impl Handler<EventMessage> for TracingActor {
+	async fn handle(&mut self, msg: EventMessage, _: &mut Context<Self>) {
+		log::info!("Event: {:?}", msg);
```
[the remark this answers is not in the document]
insipx: yep!

insipx:
Going to open this up for review. I have some more minor things to do, but other than that it's ready. Need to do some housekeeping, update schema diagram/wiki, fix some minor issues with filtering, as well as running it on Kusama/Polkadot to make sure the WASM blobs work for those chains.

Here's the latest data format:

first few columns
last few columns

substrate-archive/src/actors/workers/state_tracing.rs
```diff
@@ -0,0 +1,566 @@
...
+		Ok(())
+	}
+}
+
+#[async_trait::async_trait]
+impl<B: BlockT> Actor for TracingActor<B> {
+	async fn started(&mut self, ctx: &mut Context<Self>) {
+		let addr = ctx.address().expect("Actor just started");
+		let handler = ArchiveTraceHandler::new(addr, self.targets.clone());
+		if tracing::subscriber::set_global_default(handler).is_err() {
```
mattrutherford, line 484: It's possible to limit the scope of a subscriber to a closure, which might make it easier to group traces by block, rather than setting a global default:

https://docs.rs/tracing/0.1.22/tracing/dispatcher/fn.with_default.html

insipx:
In the refactor to use the thread local `set_default` rather than global logger:

- Removal of the state-tracing actor. Traces from the block-execution task are now sent to storage aggregator before being sent to the database. This makes sense because it allows us to collate multiple blocks of traces and batch-insert them into the database.

- The tracing subscriber is now in `wasm_tracing.rs` file at the root of `src`

- Tree data structures are removed, since we no longer need to track down root spans through all the global spans collected

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,330 @@
...
+		event.record(&mut values);
+		let parent_id =
+			event.parent().cloned().or_else(|| self.current_span.id()).ok_or(TracingError::ParentNotFound)?;
+
+		let (target, name) = if meta.name() == WASM_TRACE_IDENTIFIER {
+			match (values.0.remove(WASM_NAME_KEY), values.0.remove(WASM_TARGET_KEY)) {
+				(Some(name), Some(target)) => (name.to_string(), target.to_string()),
+				(Some(name), None) => (name.to_string(), meta.name().to_string()),
+				(None, Some(target)) => (meta.name().to_string(), target.to_string()),
+				(None, None) => (meta.name().to_string(), meta.target().to_string()),
```
dvdplm, line 136: What are these cases coming from? Is it common that substrate logs without name/target?

dvdplm (commented):
First pass, left some nits. Tracing-enabled code is a bit of a struggle to wrap my head around; but overall I think this is great stuff. I'm not sure it's as performant as it could be but we'd need to profile it to see what is slow and what isn't.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	fn enter(&self, id: &Id) {
+		self.current_span.enter(id.clone());
+	}
+
+	fn exit(&self, id: &Id) {
+		self.current_span.exit();
+		let end_time = Utc::now();
+		// let mut spans = &self.span_events.lock().spans;
```
dvdplm, line 307: dead code?

substrate-archive/src/actors.rs
```diff
@@ -198,13 +199,19 @@ where
...
+			actors.storage.clone(),
+			conf.tracing_targets.clone(),
+		);
 		let env = AssertUnwindSafe(env);
 
 		let runner = coil::Runner::builder(env, crate::TaskExecutor, &pool)
 			.register_job::<crate::tasks::execute_block::Job<B, R, C, D>>()
-			.timeout(std::time::Duration::from_secs(30))
 			.num_threads(conf.workers)
+			.timeout(Duration::from_secs(20))
```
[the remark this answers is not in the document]
insipx: Not necessarily everything -- it is a timeout for the blocks execution up until it's sent to the actor that inserts it into the database but it doesn't wait for the insert. It will try to wait for all 64 tasks to _begin_ before timing out. So, it doesn't necessarily wait until all the tasks are finished either, just that they've started to execute on the threadpool

substrate-archive/src/actors/workers/storage_aggregator.rs
```diff
@@ -51,6 +55,18 @@ where
 		}
 		Ok(())
 	}
+
+	async fn handle_traces(&mut self, ctx: &mut Context<Self>) -> Result<()> {
+		let traces = std::mem::take(&mut self.traces);
+		if !traces.is_empty() {
+			log::info!("Inserting {} traces", traces.len());
+			for trace in traces {
```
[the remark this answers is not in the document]
insipx: This is something that could be improved I think, however the reason I chose to do it this way is because each `trace` already contains probably 16-32 items (depending on what targets have been enabled) that are batch inserted, so I was worried about sending too much data at once if there were alot of traces. I think sending everything might be better though

substrate-archive/src/actors/workers/storage_aggregator.rs
```diff
@@ -66,6 +82,9 @@ where
 				if addr.send(SendStorage).await.is_err() {
 					break;
 				}
+				if addr.send(SendTraces).await.is_err() {
+					break;
```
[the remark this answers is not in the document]
insipx: They won't see anything, however our send can only fail in one way, and that's if StorageAggregator itself disconnects due to some other error: https://docs.rs/xtra/0.5.0-rc.1/xtra/address/struct.SendFuture.html

So this break/error check is to stop the task from running if the actor has shutdown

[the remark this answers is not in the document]
dvdplm: This is fine as long as users get a proper log message when postgres disconnects unexpectedly (because it was shut down or the network connection died); it sounds like you're saying this would be logged in the `StorageAggregator`, correct?

substrate-archive/src/actors.rs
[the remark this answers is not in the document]
dvdplm: Let's add a comment to that effect. :)

substrate-archive/src/actors/workers/storage_aggregator.rs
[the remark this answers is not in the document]
dvdplm: Worth investigating perhaps. If you agree, please add an issue to measure which is best.

substrate-archive/src/wasm_tracing.rs
insipx: If our traces are coming from the compiled wasm binary, then the native `name`/`target` will just be `wasm_tracing`, while the real `name`/`target` is stored in the values field of the actual trace. Due to the wasm/native barrier, native doesn't know anything about the trace except that it's a trace that's from wasm.

dvdplm: Right, so `(None, None)` is a trace from the runtime? Maybe best comment on each case to keep things clear for our future selves.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,330 @@
...
+	span_events: Arc<Mutex<SpanEvents>>,
+	targets: Vec<(String, Level)>,
+	counter: AtomicU64,
+	current_span: CurrentSpan,
+}
+
+impl TraceHandler {
+	pub fn new(targets: &str, block_num: u32, hash: BlockHash, span_events: Arc<Mutex<SpanEvents>>) -> Self {
+		let targets = targets.split(',').map(|s| parse_target(s)).collect();
+		// must start indexing from 1 otherwise `tracing` panics
```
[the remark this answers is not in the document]
insipx: The counter is used for the ID field for tracing, which panics if it's 0:
https://docs.rs/tracing/0.1.22/tracing/span/struct.Id.html#method.from_u64

Not really sure what the rationale is, maybe something internal to tracing

substrate-archive/src/actors/workers/storage_aggregator.rs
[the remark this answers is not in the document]
insipx: Yeah, if something happens with Postgres it would be logged with the database actor

Since `StorageAggregator` is mostly meant as a way to optimize our inserts by gathering bulk data, the errors which can occur are pretty minimal, so mostly anything logged here would be due to actor disconnects external to the Aggregator, which would be logged

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,330 @@
+// Copyright 2017-2019 Parity Technologies (UK) Ltd.
+// This file is part of substrate-archive.
+
+// substrate-archive is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// substrate-archive is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
```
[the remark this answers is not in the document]
insipx: I think this is fixed (the whitespace before `See` was weird, right?), also split the third line into two (as it is in the rest of the project)

substrate-archive/src/wasm_tracing.rs
insipx: Changed the block to: https://github.com/paritytech/substrate-archive/pull/134/commits/d4ef899f02fbc2ebd479c99687e4d2849e5711ed

```rust
    let name = values.0.remove(WASM_NAME_KEY).map(|t| t.to_string()).unwrap_or(meta.name().to_string());
    let target = values.0.remove(WASM_TARGET_KEY).map(|t| t.to_string()).unwrap_or(meta.target().to_string());
```

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,330 @@
...
+use substrate_archive_common::{Result, TracingError};
+use tracing::{
+	event::Event,
+	field::{Field, Visit},
+	span::{Attributes, Id, Record},
+	Level, Metadata, Subscriber,
+};
+use tracing_subscriber::CurrentSpan;
+
+type BlockHash = Vec<u8>;
```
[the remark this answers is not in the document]
insipx: Yes, we don't know the length from a Substrate block's associated type so it's a `Vec<u8>` to make room for implementations other than Polkadot

Added a comment for this

substrate-archive/src/tasks.rs
```diff
@@ -44,6 +48,10 @@ where
 	B: BlockT + Unpin,
 	B::Hash: Unpin,
 {
+	// Tracing targets
+	// if `Some` will trace the execution of the block
+	// and the targets will be sent to the TracingActor.
```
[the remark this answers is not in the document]
insipx: Yep, I just changed `TracingActor` to `StorageAggregator`

substrate-archive/src/actors/workers/blocks.rs
```diff
@@ -152,10 +152,17 @@ where
...
-			.notify_interval(std::time::Duration::from_secs(5), || Crawl)
-			.expect("Actor cannot be disconnected; just started");
-		smol::spawn(fut).detach();
+		addr.do_send(ReIndex).expect("Actor cannot be disconnected; just started");
+
+		smol::spawn(async move {
+			loop {
+				smol::Timer::after(std::time::Duration::from_secs(5));
+				if addr.send(Crawl).await.is_err() {
+					break;
```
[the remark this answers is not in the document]
insipx: After looking at this again, I don't think we need the Timer here since we await the future to finish anyway -- the purpose of the future is to just keep the blocks actor indexing.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	/// Get the block number of the block these spans come from.
+	pub fn block_num(&self) -> u32 {
+		self.block_num
+	}
+}
+
+#[derive(Debug)]
+pub struct SpanEvents {
```
mattrutherford, line 99: small nit and totally a personal preference, but I would call it `SpansAndEvents` otherwise perhaps it seems to imply events belonging to a single span

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+/// The Subscriber implementation is blocking. It uses Mutex primitives to coalesce traces before
+/// sending them to the appropriate actor.
+/// Therefore, one must be careful not to block the async executor when adding tracing spans
+/// using this subscriber implementation anywhere inside an async context in substrate-archive.
+pub struct TraceHandler {
+	/// the Block Number this trace set is from
+	block_num: u32,
+	/// Hash of the block for this trace set
+	hash: BlockHash,
+	span_events: Arc<Mutex<SpanEvents>>,
```
mattrutherford, line 114: This seems OK - but alternatively maybe put Spans and Events in separately behind their own Mutex?

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+		let counter = AtomicU64::new(1);
+		Self { targets, counter, span_events, block_num, hash, current_span: Default::default() }
+	}
+
+	/// Formats an event as an [`EventMessage`] and stores it in the [`SpanEvents`] (which is sent to the [`TracingActor`] after the block is executed).
+	fn gather_event(&self, event: &Event<'_>, time: DateTime<Utc>) -> Result<()> {
+		let meta = event.metadata();
+		let mut values = TraceData::default();
+		event.record(&mut values);
+		let parent_id =
```
mattrutherford, line 133: Is there any reason to discard events without a parent?

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+			event.parent().cloned().or_else(|| self.current_span.id()).ok_or(TracingError::ParentNotFound)?;
+
+		// check if WASM traces specify a different name/target.
```
mattrutherford, line 136: Just curious as to the proportion of Spans and Events coming from WASM vs native code in archive?

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	fn record_bool(&mut self, field: &Field, value: bool) {
+		self.0.insert(field.name().to_string(), DataType::Bool(value));
+	}
+}
+
+impl Subscriber for TraceHandler {
+	fn enabled(&self, metadata: &Metadata<'_>) -> bool {
+		self.targets.iter().any(|t| metadata.target().starts_with(t.0.as_str()))
```
mattrutherford, line 254: maybe `self.targets.iter().any(|(t,_l)| metadata.target().starts_with(t.as_str()))`

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+		attrs.record(&mut values);
+
+		let id = Id::from_u64(self.counter.fetch_add(1, Ordering::Relaxed));
+		let span_message = SpanMessage {
+			id: id.clone(),
+			parent_id: attrs.parent().cloned().or_else(|| self.current_span.id()),
+			name: meta.name().to_string(),
+			target: meta.target().to_string(),
+			level: *meta.level(),
+			start_time: Utc::now(),
```
mattrutherford, line 269: A different approach could be to use start time (measured in ns) relative to the beginning of execution of the current block, using the monotonic `std::time::Instant` - in which case start time would be the duration between the block start and the current time. Duration would be difference between this time and the exit/close time. Perhaps this is more relevant for later analysis, but probably comes down to personal choice, so I'd be ok with either.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	fn record(&self, id: &Id, values: &Record<'_>) {
+		if let Some(span) = self.span_events.lock().spans.iter_mut().find(|span| &span.id == id) {
+			values.record(&mut span.values);
+		}
+	}
+
+	fn record_follows_from(&self, _: &Id, _: &Id) {
+		log::warn!("Followed span relationship not recorded");
```
mattrutherford, line 290: We don't currently make use of this in substrate in any of the implementations, but maybe as a follow up PR we can add a `follows_from_id` field to `SpanMessage` in case it's used in future.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+		if let Err(e) = self.gather_event(event, time) {
+			log::error!("{}", e.to_string());
+		}
+	}
+
+	fn enter(&self, id: &Id) {
+		self.current_span.enter(id.clone());
+	}
+
+	fn exit(&self, id: &Id) {
```
mattrutherford, line 304: The tracing spec does not guarantee that once a span has exitied that it will not be entered again, so perhaps this code could be moved to `try_close`

mattrutherford (commented):
Overall looks good - just a small query about Events without parent

substrate-archive/src/wasm_tracing.rs
insipx: Mostly it will be WASM trace up until the syncing gets to the native runtime. For example, on Kusama it would be wasm from block 3.2ml up until 5.8 or something (roughly 8.24 - 26 I think), and then it would recognize version 0.8.27 as native and switch to that

substrate-archive/src/wasm_tracing.rs
insipx: Good catch! This was left over from when I wrapped the entire block execution in a span, so anything outside of that span I considered irrelevant. Now that the subscriber is thread-local that's no longer needed

substrate-archive/src/database.rs
```diff
@@ -269,6 +271,117 @@ impl Insert for Metadata {
...
+		Err(ArchiveError::TimestampOutOfRange)
+	} else {
+		Ok(time.to_std().expect("Checked for less than 0"))
+	}
+}
+
+// the alternative to this is creating a Composite postgres type
+// that stores seconds + extra nanoseconds.
+// Postgres does not support nanosecond precision.
+fn shave_nanos(time: std::time::Duration) -> Result<std::time::Duration> {
```
mattrutherford, line 380: what about casting `Duration::as_nanos()` from `u128` to `i64` ? then could be stored in a `bigint`? Still allows for 292 years duration at nanosecond precision.

insipx: Could definitely do that -- I didn't do it the first time around because I liked that it was in an 'official' Postgres time format. However if it's just nanos that's fairly simple to convert back and forth

mattrutherford: yeah, if nanos aren't important then using interval def has some advantages - there's also `Duration::as_secs_f64` which would avoid a cast

substrate-archive/src/wasm_tracing.rs
insipx: https://github.com/paritytech/substrate-archive/issues/201

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,332 @@
...
+	fn exit(&self, _id: &Id) {
+		self.current_span.exit();
+	}
+
+	fn try_close(&self, id: Id) -> bool {
+		let end_time = Utc::now();
+		if let Some(span) = self.span_events.lock().spans.iter_mut().find(|span| &span.id == &id) {
+			span.overall_time = end_time - span.start_time;
+		}
+		// try_close returns false by default -- we want to keep this behavior.
```
insipx, line 312: Not actually sure about this -- we don't track span references by id or anything, so I figured that doing what the default impl does is the most sensible thing here. It works on my end for the DB. But not sure if returning `false` unilaterally here means that we can re-enter the span again sort of like in `exit`. @mattrutherford ?

by using this does this mean we have to also track clone_span to make sure all spans are dropped?

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+/// The Event a tracing subscriber collects before sending data to the TracingActor.
+#[derive(Debug)]
+pub struct EventMessage {
+	pub block_num: Option<u32>,
+	pub hash: Option<Vec<u8>>,
+	pub name: String,
+	pub target: String,
+	pub level: Level,
+	pub values: TraceData,
+	pub parent_id: Id,
```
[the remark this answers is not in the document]
insipx: this is now an Option<Id> since we stopped throwing out Events w/o a parent id

substrate-archive/src/wasm_tracing.rs
insipx: Yeah -- not sure if there would be much benefit since the runtime is single threaded anyway so we shouldn't hit this mutex iin different threads at the same time very much; it's cleaner to instantiate this way imo (this is mostly personal preference though. If it is actually better to use two mutexes I can do that)

substrate-archive/src/wasm_tracing.rs
insipx: Not sure I fully understand the benefit of this over just using Utc::now() -- would this give us more precise/extra information over when block execution started relative to the rest of the execution?

substrate-archive/src/wasm_tracing.rs
dvdplm: There are many interesting performance questions around this PR and I think we could run a good number of experiments that are interesting to the wider substrate team. At this time though I'm more interested in delivering the state-tracing feature and so I'm deliberately holding back the itch to benchmark stuff.

So yeah, I'm fine keeping a single mutex for now. :)

substrate-archive/src/wasm_tracing.rs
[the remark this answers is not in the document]
dvdplm: I went looking briefly in the code and it's more complex than I thought. Didn't quite understand what the reason is. Oh well.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,329 @@
...
+
+		self.targets.iter().filter(|t| t.0.as_str() != "wasm_tracing").any(|t| {
+			let check_target = |target: &str, lvl: Level| -> bool { target.starts_with(&t.0.as_str()) && lvl <= t.1 };
+
+			check_target(&span.target, span.level)
+				|| wasm_target.as_ref().map(|wasm_t| check_target(wasm_t, span.level)).unwrap_or(false)
+		})
+	}
+
+	/// Formats spans based upon data types that would be more useful for querying in the context
```
dvdplm, line 176:
```suggestion
	/// Formats spans based upon data types that are more useful for querying in the context
```

substrate-archive/src/archive.rs
```diff
@@ -137,6 +149,20 @@ impl<B, R, D, DB> Builder<B, R, D, DB> {
 		self.max_block_load = Some(max_block_load);
 		self
 	}
+
+	/// Set the folder and targets for tracing.
+	/// This tells substrate-archive to also store all state-traces resulting from the execution of blocks.
+	///
+	/// # Note
+	/// Traces will only be collected if a coexisting WASM binary
+	/// for the runtime version of the block being currently executed is available.
```
dvdplm, line 158: Oh, I realize I must have missed an important piece here. I thought that we were collecting traces in native mode (and that this was the "easy" part), but that's not correct at all then? Hmm. Does this mean that as soon as we have a "hole" in the chain for which we do not have a WASM blob we don't get any state traces at all?

substrate-archive/src/database.rs
dvdplm: There's some tension here between making the data easy to use (native PG time format is good here) and high precision. As the primary use case for the time being is balance reconciliation I'd say we stick with the native time format for now and perhaps follow up by adding a new column with the full-precision timestamp (or add it right now if it's easy to do)?

substrate-archive/src/archive.rs
insipx: Well native execution is a part of it, so even if the runtime that you're running does not have a WASM binary then tracing will work. But yeah there will be holes in traces if a wasm binary isn't present for X version. So far that means anything older than 8.25

dvdplm: Printing out what runtimes we know of seems like a good QoL thing.
And a big, visible, log message when we can't collect anymore traces? Or even halt?

But why can't `archive` switch to the new runtime and chug along?

substrate-archive/src/wasm_tracing.rs
insipx: I think at some point it would also be interesting to benchmark other parts of archive too, and optimize some of our defaults (block load/threads/wasm cache etc)

substrate-archive/src/archive.rs
insipx: The wasm binaries that have tracing enabled are kept in the repo, so whenever there's an upgrade I would have to compile that binary with the new version, push a commit, and then those running archive would need to upgrade to the newer version/pull the binary from the repo in order to have traces work again. By default substrate does not have wasm tracing-enabled binaries (the feature does not even officially exist in polkadot, it requires manual compilation).

dvdplm: Ofc, I'm dumb, thank you for explaining (again).

We'll have to think about how we can make this process a bit less painful, perhaps we can ask @s3krit to build extra artifacts for us as part of the normal release pipeline and we can get "official" builds with wasm tracing enabled at some point. Then we can either script pulling them into this repo or even point directly at our releases.

dvdplm: > Printing out what runtimes we know of seems like a good QoL thing.
> And a big, visible, log message when we can't collect anymore traces? Or even halt?

Can we capture ^^^ in issues so we don't stall this PR any longer?

dvdplm:
CI's angry.

dvdplm (approved):
I think this is good to go (modulo CI).

bin/node-template-archive/src/cli_opts.rs
```diff
@@ -37,12 +37,36 @@ impl CliOpts {
 			3 => log::LevelFilter::Debug,
 			4 | _ => log::LevelFilter::Trace,
 		};
+<<<<<<< HEAD
```
maciejhirsz, line 40: Merge conflict?

substrate-archive-backend/src/block_exec.rs
```diff
@@ -108,8 +108,8 @@ where
 
 		let state = self.backend.state_at(self.id)?;
 
-		// FIXME: For some reason, wasm runtime calculates a different number of digest items
-		// then what we have in the block
+		// Wasm runtime calculates a different number of digest items
+		// than what we have in the block
```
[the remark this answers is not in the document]
maciejhirsz: Maybe because wasm target is 32bit instead of 64?

bin/node-template-archive/src/cli_opts.rs
insipx: Yeah, somehow missed it, fixed

substrate-archive/src/archive.rs
insipx: immortalized in #207

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,332 @@
...
+		Ok(())
+	}
+
+	// we need this because we don't know the values until after tracing has been executed
+	/// Returns true if a span is part of an enabled Target. Checks WASM in addition to the spans target.
+	fn is_enabled(&self, span: &SpanMessage) -> bool {
+		let wasm_target = span.values.0.get(WASM_TARGET_KEY).map(|s| s.to_string());
+
+		self.targets.iter().filter(|t| t.0.as_str() != "wasm_tracing").any(|t| {
+			let check_target = |target: &str, lvl: Level| -> bool { target.starts_with(&t.0.as_str()) && lvl <= t.1 };
```
maciejhirsz, line 169: This looks like an overkill, why not inline the check, or make it into a bool?

```rust
let valid_target = target.starts_with(&t.0.as_str()) && lvl <= t.1;

valid_target || wasm_target.as_ref().map(|wasm_t| check_target(wasm_t, span.level)).unwrap_or(false)
```

maciejhirsz (commented):
I really like how readable all of this code is. Some grumbles and questions.

substrate-archive/src/wasm_tracing.rs
insipx: Yeahp, I  made it simpler. Hopefully easier to understand now too

substrate-archive/src/wasm_tracing.rs
```diff
@@ -0,0 +1,332 @@
...
+	hash: BlockHash,
+	span_events: Arc<Mutex<SpansAndEvents>>,
+	targets: Vec<(String, Level)>,
+	counter: AtomicU64,
+	current_span: CurrentSpan,
+}
+
+impl TraceHandler {
+	pub fn new(targets: &str, block_num: u32, hash: BlockHash, span_events: Arc<Mutex<SpansAndEvents>>) -> Self {
+		let targets = targets.split(',').map(|s| parse_target(s)).collect();
```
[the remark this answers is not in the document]
insipx: Yeah -- giving this a lifetime would be complicated and probably require some refactoring to `coil` and how it handles the Environment state that's passed around to different threads. I believe it might be possible but would involve lots of complicated 'static (I just tried this out). I think I may revisit this later, though.

substrate-archive/src/actors/workers/storage_aggregator.rs
```diff
@@ -51,6 +53,18 @@ where
 		}
 		Ok(())
 	}
+
+	async fn handle_traces(&mut self, ctx: &mut Context<Self>) -> Result<()> {
+		let traces = std::mem::take(&mut self.traces);
```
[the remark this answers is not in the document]
insipx: I went with the second option (might as well avoid allocations, right? :P)

I'm pretty sure this is right: 

```rust
let mut traces = std::mem::take(&mut self.traces);
		if !traces.is_empty() {
			log::info!("Inserting {} traces", traces.len());
			for trace in traces.drain(..) {
				let send_result = self.db.send(trace.into()).await?;
				ctx.handle_while(self, send_result).await;
			}
		}
		std::mem::swap(&mut self.traces, &mut traces);
```

bin/polkadot-archive/Cargo.toml
```diff
@@ -1,20 +1,22 @@
 [package]
 name = "polkadot-archive"
-version = "0.2.2"
+version = "0.2.3"
 authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
 edition = "2018"
 
 [dependencies]
+polkadot-service = { git = "https://github.com/paritytech/polkadot", branch = "master" }
+frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
niklasad1, line 9: looks not to be used, I guess it's fetched from other dependencies but still....

```suggestion
```

substrate-archive-backend/Cargo.toml
```diff
@@ -22,7 +22,7 @@ parity-util-mem = "0.8"
 
 # Substrate
 sc-client-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
-sc-executor = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sc-executor = { git = "https://github.com/paritytech/substrate", branch = "master", features = ["wasmtime"] }
```
niklasad1, line 25: `wasmtime` only works for `x64` last I checked,

[substrate does this for example](https://github.com/paritytech/substrate/blob/master/bin/node/cli/Cargo.toml#L108-#L112)

You don't have to fix it in this PR but we should file an issue because most likely someone is going to run into this problem.

substrate-archive-backend/src/frontend.rs
```diff
@@ -58,11 +80,12 @@ where
 {
 	let backend = Arc::new(ReadOnlyBackend::new(db, true));
 
-	let executor =
-		NativeExecutor::<Dispatch>::new(WasmExecutionMethod::Interpreted, Some(wasm_pages), block_workers as usize);
-
-	let executor =
-		LocalCallExecutor::new(backend.clone(), executor, Box::new(TaskExecutor::new()), Default::default())?;
+	let executor = NativeExecutor::<Dispatch>::new(
+		WasmExecutionMethod::Interpreted,
```
niklasad1, line 84: why not `WasmExecutionMethod::Compiled`? It's enabled in this PR and I suppose it's faster?!

substrate-archive/src/actors/workers/storage_aggregator.rs
```diff
@@ -51,6 +53,19 @@ where
...
+	async fn handle_traces(&mut self, ctx: &mut Context<Self>) -> Result<()> {
+		let mut traces = std::mem::take(&mut self.traces);
+		if !traces.is_empty() {
+			log::info!("Inserting {} traces", traces.len());
+			for trace in traces.drain(..) {
+				let send_result = self.db.send(trace.into()).await?;
+				ctx.handle_while(self, send_result).await;
+			}
+		}
+		std::mem::swap(&mut self.traces, &mut traces);
```
niklasad1, line 66: I don't understand why `swap` is performed here,  `traces` should be empty anyways after `drain(..)`

substrate-archive/src/actors/workers/storage_aggregator.rs
```diff
@@ -51,6 +53,19 @@ where
 		}
 		Ok(())
 	}
+
+	async fn handle_traces(&mut self, ctx: &mut Context<Self>) -> Result<()> {
+		let mut traces = std::mem::take(&mut self.traces);
```
niklasad1, line 58: why not `let mut traces = std::mem::replace(&mut self.traces, Vec::with_capacity(250));` as it's done for the storage?

I have no clue about why 250 is used but just raising the question because traces and storage are handled different ^^

substrate-archive/src/archive.rs
```diff
@@ -44,6 +44,12 @@ const POSTGRES_VAR: &str = "DATABASE_URL";
 /// The recommended open file descriptor limit to be configured for the process.
 const RECOMMENDED_OPEN_FILE_DESCRIPTOR_LIMIT: u64 = 10_000;
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
niklasad1, line 47: maybe some docs? :)

substrate-archive/src/actors/workers/storage_aggregator.rs
maciejhirsz: This is form my review. It keeps the capacity of the vector (`drain` removes the items, but it doesn't deallocate) without having to reallocate a new `Vec` when you take ownership of it.

maciejhirsz (approved):

substrate-archive-backend/src/frontend.rs
dvdplm: I measured this and it's actually not faster. I was surprised by this and it might be worth taking a deeper look. When syncing the chain in normal `polkadot` using `--wasm-execution=Compiled` is much much faster.

substrate-archive/src/database.rs
```diff
@@ -272,3 +275,111 @@ impl Insert for Metadata {
...
+				block_num, hash, is_event, timestamp, duration, file, line, trace_id, trace_parent_id, target, name, traces
+			) VALUES
+			"#,
+			r#"
+			ON CONFLICT DO NOTHING
+			"#,
+		);
+
+		let block_num = self.block_num();
+		let hash = self.hash().to_vec();
```
niklasad1, line 296: The hash is cloned here because `self.spans` and `self.events` are moved?

I guess the hash small so probably fine.

substrate-archive/src/database.rs
```diff
@@ -272,3 +275,111 @@ impl Insert for Metadata {
...
+		for span in self.spans.into_iter() {
+			let id = i32::try_from(span.id.into_u64())?;
+			let parent_id: Option<i32> =
+				if let Some(id) = span.parent_id { Some(i32::try_from(id.into_u64())?) } else { None };
+			let overall_time: i64 = time_to_std(span.overall_time)?.as_nanos().try_into()?;
+			batch.reserve(12)?;
+			if batch.current_num_arguments() > 0 {
+				batch.append(",");
+			}
+			batch.append("(");
```
niklasad1, line 307: This is hard to review, would be cool with a helper function/macro or something in the future for handle `bind` and `append`

substrate-archive/src/database.rs
```diff
@@ -272,3 +275,111 @@ impl Insert for Metadata {
...
+// Old time is disabled in chrono by not providing the feature flag in Cargo.toml.
+fn time_to_std(time: chrono::Duration) -> Result<std::time::Duration> {
+	if time < chrono::Duration::zero() {
+		Err(ArchiveError::TimestampOutOfRange)
+	} else {
+		Ok(time.to_std().expect("Checked for less than 0"))
+	}
+}
+
+#[cfg(test)]
```
niklasad1, line 382: remove?

substrate-archive/src/database.rs
```diff
@@ -272,3 +275,111 @@ impl Insert for Metadata {
...
+
+		Ok(batch.execute(conn).await?)
+	}
+}
+
+// Chrono depends on an error type in `time` that is a full version behind the one that SQLX uses
+// This function avoids depending on two time lib.
+// Old time is disabled in chrono by not providing the feature flag in Cargo.toml.
+fn time_to_std(time: chrono::Duration) -> Result<std::time::Duration> {
+	if time < chrono::Duration::zero() {
```
niklasad1, line 375: why not `time.to_std().map_err(|_| ArchiveError::TimestampOutOfRange)` instead of expect!?

substrate-archive-backend/Cargo.toml
insipx: Didn't know this actually! By default we run in interpreted which shouldn't use wasmtime, but this is good to know. Created #209 and #210 for this

substrate-archive/src/tasks.rs
```diff
@@ -90,16 +104,39 @@ where
...
+
+		let _guard = handler.map(tracing::subscriber::set_default);
+
+		let now = std::time::Instant::now();
+		let block = BlockExecutor::new(api, &env.backend, block)?.block_into_storage()?;
+		log::debug!("Took {:?} to execute block", now.elapsed());
+
+		Storage::from(block)
+	};
+	let traces = Arc::try_unwrap(span_events).map_err(|_| TracingError::NoTraceAccess)?.into_inner();
```
niklasad1, line 132: > https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap
Returns the inner value, if the Arc has exactly one strong reference.

So I guess that this relies on that `the storage block` will be executed first increase the `Arc strong count to 2` and then when it's finished it will  drop the `Arc` and decrease the strong count 1 and then you destroy the Mutex because you know that you it's not aliased/mutual exlusion anymore?

This is not super obvious to me and deserves a comment. `Arc::get_mut` and `Arc::try_unwrap` should be used at little as possible IMHO.

substrate-archive/src/database.rs
insipx: Yeah, definitely something that needs to be done and would offer lots of QoL improvements over the current batch-insertion strategy. I opened #211 to track this w/ a bit of history around it. Definitely redoing the batch-insertion logic is on my list of todo's though

substrate-archive/src/actors/workers/storage_aggregator.rs
niklasad1: ok, I see.

substrate-archive-backend/src/frontend.rs
niklasad1: Ok, I see.

niklasad1 (approved):

substrate-archive/src/tasks.rs
insipx: An alternative is not destroying the Mutex and draining the Vec's inside of it into two new Vector's, which avoids all this unwrapping logic. That introduces two new allocations though, and since this block execution stuff needs to churn through  millions of blocks (with 1000 - 100K loaded at once) I chose to unwrap the mutex (which I _think_ avoids new mem alloc). Archive is using lots of memory as-is with what it's loading, and the mutex is used here in fairly limited scope (famous last words, though).

Another interesting performance question for #205 as to whether this actually helps

Adding a comment explaining it though :+1:

substrate-archive/src/database.rs
insipx: actually was able to make it all references

substrate-archive/src/database.rs
insipx: i don't remember why I originally did it with `if` but the map_err is definitely better, fixed

PR #141, opened 2021-01-05
Update xtra to v0.5.0-rc.1

koushiro:

* update xtra to v0.5.0-rc.1
* reorder Cargo.toml and imports
* apply rustfmt for all crates

### DIFF
--- a/bin/node-template-archive/src/cli_opts.rs
+++ b/bin/node-template-archive/src/cli_opts.rs
@@ -14,53 +14,52 @@
 // You should have received a copy of the GNU General Public License
 // along with substrate-archive.  If not, see <http://www.gnu.org/licenses/>.
 
-use clap::{load_yaml, App};
 use std::path::PathBuf;
+
+use clap::{load_yaml, App};
 
 #[derive(Clone)]
 pub struct CliOpts {
-    pub file: PathBuf,
-    pub log_level: log::LevelFilter,
-    pub chain_spec: node_template::chain_spec::ChainSpec,
+	pub file: PathBuf,
+	pub log_level: log::LevelFilter,
+	pub chain_spec: node_template::chain_spec::ChainSpec,
 }
 
 impl CliOpts {
-    pub fn parse() -> Self {
-        let yaml = load_yaml!("cli_opts.yaml");
-        let matches = App::from(yaml).get_matches();
-        let log_level = match matches.occurrences_of("verbose") {
-            0 => log::LevelFilter::Info,
-            1 => log::LevelFilter::Info,
-            2 => log::LevelFilter::Info,
-            3 => log::LevelFilter::Debug,
-            4 | _ => log::LevelFilter::Trace,
-        };
-        let file = matches
-            .value_of("config")
-            .expect("Config is a required value");
-        let chain_spec;
-        let spec = matches.value_of("spec");
-        if spec.is_some() {
-            match spec {
-                Some("dev") => {
-                    chain_spec = node_template::chain_spec::development_config();
-                }
-                Some("") | Some("local") => {
-                    chain_spec = node_template::chain_spec::local_testnet_config();
-                }
-                path => {
-                    chain_spec = node_template::chain_spec::ChainSpec::from_json_file(
-                        std::path::PathBuf::from(path.expect("checked for existence; qed")),
-                    )
-                }
-            }
-        } else {
-            panic!("Chain spec could not be loaded; is the path correct?")
-        }
-        CliOpts {
-            file: PathBuf::from(file),
-            log_level,
-            chain_spec: chain_spec.expect("Chain spec could not be loaded"),
-        }
-    }
+	pub fn parse() -> Self {
+		let yaml = load_yaml!("cli_opts.yaml");
+		let matches = App::from(yaml).get_matches();
+		let log_level = match matches.occurrences_of("verbose") {
+			0 => log::LevelFilter::Info,
+			1 => log::LevelFilter::Info,
+			2 => log::LevelFilter::Info,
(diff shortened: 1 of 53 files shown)


koushiro:
@insipx PTAL

substrate-archive-backend/Cargo.toml
```diff
@@ -6,38 +6,36 @@ edition = "2018"
...
-sp-blockchain = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-client-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sc-executor = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sc-service = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-block-builder = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-blockchain = { git = "https://github.com/paritytech/substrate", branch = "master" }
 sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-database = { git = "https://github.com/paritytech/substrate", branch = "master" }
+#sp-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
niklasad1, line 32: remove?

bin/node-template-archive/src/cli_opts.rs
```diff
@@ -14,53 +14,52 @@
...
 use std::path::PathBuf;
 
+use clap::{load_yaml, App};
+
 #[derive(Clone)]
 pub struct CliOpts {
-    pub file: PathBuf,
-    pub log_level: log::LevelFilter,
-    pub chain_spec: node_template::chain_spec::ChainSpec,
+	pub file: PathBuf,
```
niklasad1, line 23: please use the [.editconfig](https://github.com/paritytech/substrate-archive/blob/master/.editorconfig) for `rustfmt`, we use `tabs` instead of spaces for example.

niklasad1 (commented):
The motivation for bumping `xtra` to `v0.5.0-rc.1` is unclear to me, can you please explain why this is useful?

bin/node-template-archive/src/cli_opts.rs
koushiro: Yeah, I know. Actually, the original indent is space, and the modified one is tab

niklasad1: Ok, sorry my bad. These may have been missed because they are excluded for the workspace.

koushiro:
@niklasad1 If nothing goes wrong, the author of `xtra` will release version 0.5.0 in two weeks, which will not change much compared to the `0.5.0-rc.1`. But `substrate-archive` is using `xtra 0.5.0-beta.4` now that is a version released half a year ago and differs greatly from the current `xtra`.

insipx (commented):
very minor nits but LGTM. I think keeping up to date with xtra's release and updating smol is good

There are a few places that introduce spaces between imports that I think should be removed. There's the one in `test_utils` but in other files too.

I was able to run this branch against Polkadot 0.8.26-1 successfully

koushiro:
@insipx I prefer to distinguish imports based on the source of dependencies. If you don't add a blank line between imports from different sources, the `rustfmt` will merge them :(

substrate-archive-common/src/error.rs
```diff
@@ -16,7 +16,7 @@
 use std::{env, io};
 use thiserror::Error;
 
-pub type Result<T> = std::result::Result<T, Error>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
niklasad1, line 19: This change doesn't make sense to me, either make it specific to this module or use the generic Result from the standard library.

//cc @insipx

substrate-archive/Cargo.toml
```diff
@@ -6,61 +6,45 @@ edition = "2018"
 
 [dependencies]
 # external
+async-trait = "0.1"
+coil = { git = "https://github.com/insipx/coil", package = "coil" }
```
niklasad1, line 10: needless `package name`

```suggestion
coil = { git = "https://github.com/insipx/coil" }
```

koushiro:
@insipx I prefer to use `mod.rs` instead of `foo.rs and foo/` for module management (just my taste), if you don’t mind, I’ll change it.

niklasad1:
> I prefer to use mod.rs instead of foo.rs and foo/ for module management (just my taste), if you don’t mind, I’ll change it.

@koushiro Please don't do it in this PR, it's large enough as it is. Let Andrew review one more time and I think it's "mergable" as it is.

If you have strong opinions on it, please do it another PR.

koushiro:
@niklasad1 okay

substrate-archive-common/src/error.rs
insipx: I don't believe it to be strictly necessary either, I just see it as an alternative way to represent the `Result` that was already being used.

That being said, I don't see anything wrong with it perse, and don't have strong opinions either way

insipx (approved):

insipx:
> @insipx I prefer to use `mod.rs` instead of `foo.rs and foo/` for module management (just my taste), if you don’t mind, I’ll change it.

As @niklasad1 mentioned, this would definitely be `make another PR` material. That being said, I would prefer to stick with the `foo/` `foo.rs` structure we already have. I like not having lots of `mod.rs`, since it can be confusing to quickly grep through the files and differentiate one `mod.rs` from another. I am open to hearing arguments for refactoring our codebase to mod.rs, I just don't think it's too useful tbh.

niklasad1 (approved):
thank you.

PR #142, opened 2021-01-07
Improve SqlBlockBuilder

koushiro:
~~waiting for #141 to be merged firstly~~

### DIFF
--- a/substrate-archive/src/sql_block_builder.rs
+++ b/substrate-archive/src/sql_block_builder.rs
@@ -23,85 +23,55 @@ use std::marker::PhantomData;
 use codec::{Decode, Encode};
 use sp_runtime::{
 	generic::SignedBlock,
-	traits::{Block as BlockT, DigestFor, Header as HeaderT},
+	traits::{Block as BlockT, Header as HeaderT},
 };
 
 use substrate_archive_common::{models::BlockModel, types, Error};
 
-struct GenericParts<B: BlockT>(
-	<B::Header as HeaderT>::Hash,
-	<B::Header as HeaderT>::Hash,
-	<B::Header as HeaderT>::Hash,
-);
-
-pub struct BlockBuilder<B: BlockT> {
+pub struct SqlBlockBuilder<B: BlockT> {
 	_marker: PhantomData<B>,
 }
 
-impl<'a, B: BlockT> BlockBuilder<B> {
+impl<'a, B: BlockT> SqlBlockBuilder<B> {
 	/// With a vector of BlockModel
 	pub fn with_vec(blocks: Vec<BlockModel>) -> Result<Vec<types::Block<B>>, Error> {
 		blocks
 			.into_iter()
 			.map(|b| {
-				let (b, s) = Self::with_single(b)?;
-				let b = SignedBlock { block: b, justification: None };
-				Ok(types::Block::new(b, s))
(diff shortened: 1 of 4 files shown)


insipx:
nice! This certainly makes sql block builder more concise.

Just wondering why move `sql_block_model` to `actors/`? It's not apart of the `xtra`/actor flow so i'd think it belongs in `src/`root

koushiro:
Because the `sql_block_builder` is only used in the `actors` module, if you don’t like it, I'll revert it

koushiro:
@insipx @niklasad1 PTAL

insipx (approved):

substrate-archive/src/sql_block_builder.rs
```diff
@@ -23,85 +23,55 @@ use std::marker::PhantomData;
...
-		Ok(GenericParts(
-			Decode::decode(&mut parent_hash)?,
-			Decode::decode(&mut state_root)?,
-			Decode::decode(&mut extrinsics_root)?,
-		))
-	}
+struct BlockDecoder<B: BlockT> {
+	header: B::Header,
+	ext: Vec<B::Extrinsic>,
+	spec: u32,
```
niklasad1, line 59: what's the reason why `BlockModel` uses i32 as spec and `BlockDecoder` uses u32?

insipx: `BlockMode`l is the result of a query from Postgres, and since Postgres does not support unsigned integers, only the signed i32/i64 rust types implement `Encode`/`Decode` in SQLx.

niklasad1 (approved):

substrate-archive/src/sql_block_builder.rs
```diff
@@ -23,85 +23,55 @@ use std::marker::PhantomData;
...
 	/// With a vector of BlockModel
 	pub fn with_vec(blocks: Vec<BlockModel>) -> Result<Vec<types::Block<B>>, Error> {
 		blocks
 			.into_iter()
 			.map(|b| {
-				let (b, s) = Self::with_single(b)?;
-				let b = SignedBlock { block: b, justification: None };
-				Ok(types::Block::new(b, s))
+				let (block, spec) = Self::with_single(b)?;
+				let block = SignedBlock { block, justification: None };
```
emostov, line 42: Not related to this PR, but just out of curiosity is there any reason we don't store a justification? From my understanding there is at least 1 block with a stored justification per session.

emostov (approved):
Looks good to me. Thank you for the contributions! Looks like it just needs some conflicts resolved.

substrate-archive/src/sql_block_builder.rs
insipx: Here, the justification is unnecessary since if we already have the block in the database, a justification isn't needed for re-executing the block. Justification is used for consensus purposes/proving that X block is indeed finalized because some number of people/validators voted on it I think (my understanding there is rudimentary I think). Since archive doesn't do anything with consensus we can safely ignore this

However, we actually don't store justifications at all in the database. We have the justification on the block before it's inserted into the database, there's just not justification column. I can't remember if there was a reason behind this, but I think it would be good to support storing justifications.

insipx: Created an issue for this: https://github.com/paritytech/substrate-archive/issues/147

PR #144, opened 2021-01-08
Fix format checking in CI

koushiro:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,8 +2,6 @@ name: Rust
 
 on: [push, pull_request]
 
-
-
 jobs:
   check:
     name: Check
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -43,7 +41,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: fmt
-          args: --all
+          args: --all -- --check
```
niklasad1, line 44: Can you please add so that we run it on the crates excluded from the workspace?

koushiro: Done

insipx (approved):

niklasad1 (approved):

PR #160, opened 2021-01-11
Update substrate to 2.0.1

dvdplm:
Brings a bunch of other updates, `kvdb`, `kvdb-rocksdb`, `parity-util-mem` and friends.

Biggest point of uncertainty is the errors, `BlockchainError` lost the `Msg` catch-all member so it takes some creativity to match up what we have here to what's available. :/

### DIFF
--- a/substrate-archive/src/database/listener.rs
+++ b/substrate-archive/src/database/listener.rs
@@ -20,6 +20,7 @@
 //! listen wakeup.
 
 use std::pin::Pin;
+use std::time::Duration;
 
 use futures::{Future, FutureExt, StreamExt};
 use serde::{Deserialize, Serialize};
@@ -149,9 +150,9 @@ where
 			}
 			// collect the rest of the results, before exiting, as long as the collection completes
 			// in a reasonable amount of time
-			let timeout = smol::Timer::after(std::time::Duration::from_secs(1));
+			let timeout = smol::Timer::after(Duration::from_secs(1));
 			futures::select! {
-				_ = timeout.fuse() => {},
+				_ = FutureExt::fuse(timeout) => {},
 				notifs = listener.collect::<Vec<_>>().fuse() => {
 					for msg in notifs {
 						self.handle_listen_event(msg.unwrap(), &mut conn).await;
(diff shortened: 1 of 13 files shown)


emostov (approved):
lgtm

> Biggest point of uncertainty is the errors, BlockchainError lost the Msg catch-all member so it takes some creativity to match up what we have here to what's available. :/

Updated errors look reasonable to me. Should we note the error update as a breaking change so we can bump relevant versioning on next release?

Also whats up with the CI, does it need fmt again?

niklasad1:
> Also whats up with the CI, does it need fmt again?

No, it looks like we are using two different versions of `parity-util-mem`. `sc-service` requires 0.7.x and `substrate-archive-backend` is using `0.8`.

I think it goes away once you downgrade the substrate dependencies in `substrate-archive-backend` to `2.0.1` @dvdplm :)

dvdplm:
> I think it goes away once you downgrade the substrate dependencies in `substrate-archive-backend`, `substrate-archive-common` and `substrate-archive` to `2.0.1` @dvdplm :)

You surely mean upgrade? The only way is up!

(Yes, you are totally correct; I missed that the binaries are not in the workspace, ty.)

insipx (approved):

substrate-archive/src/database/listener.rs
```diff
@@ -149,9 +150,9 @@ where
 			}
 			// collect the rest of the results, before exiting, as long as the collection completes
 			// in a reasonable amount of time
-			let timeout = smol::Timer::after(std::time::Duration::from_secs(1));
+			let timeout = smol::Timer::after(Duration::from_secs(1));
 			futures::select! {
-				_ = timeout.fuse() => {},
+				_ = FutureExt::fuse(timeout) => {},
```
niklasad1, line 155: What's the reason behind this change?

insipx: iirc won't compile because it complains that .fuse() might point to the `StreamExt`, `FutureExt` or other `fuse()` method

dvdplm: Indeed. But the deeper reason the compiler can't see this escapes me.

PR #166, opened 2021-01-12
Tweaks to RuntimeVersionCache

dvdplm:
Did some benchmarking and general tinkering and tweaked a few minor things while at it.

Builds on #160 so best look at that first.

### DIFF
--- a/substrate-archive-backend/src/runtime_version_cache.rs
+++ b/substrate-archive-backend/src/runtime_version_cache.rs
@@ -87,26 +87,26 @@ impl<B: BlockT, D: ReadOnlyDB + 'static> RuntimeVersionCache<B, D> {
 		if self.versions.load().contains_key(&code_hash) {
 			Ok(self.versions.load().get(&code_hash).cloned())
 		} else {
-			log::debug!("new code hash: {:#X?}", code_hash);
-			let mut ext: BasicExternalities = BasicExternalities::default();
+			log::debug!("Adding new runtime code hash to cache: {:#X?}", code_hash);
+			let mut ext = BasicExternalities::default();
 			ext.register_extension(CallInWasmExt::new(self.exec.clone()));
-			let v: RuntimeVersion = ext.execute_with(|| {
+			let version: RuntimeVersion = ext.execute_with(|| {
 				let ver = sp_io::misc::runtime_version(&code).ok_or(ArchiveError::WasmExecutionError)?;
 				decode_version(ver.as_slice())
 			})?;
-			log::debug!("Registered New Runtime Version: {:?}", v);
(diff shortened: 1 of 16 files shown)


insipx (approved):
cool, LGTM!

substrate-archive-backend/src/runtime_version_cache.rs
```diff
@@ -120,12 +120,12 @@ impl<B: BlockT, D: ReadOnlyDB + 'static> RuntimeVersionCache<B, D> {
 	where
 		NumberFor<B>: Into<u32>,
 	{
-		let versions = self.find_versions(blocks.as_slice())?;
+		let versions = self.find_versions(&blocks)?;
```
emostov, line 123: dq: are these exactly equivalent?

emostov (approved):
lgtm modulo a nit I am not sure about concerning `RangeBounds` vs using boolean logic (see comment)

substrate-archive-backend/src/runtime_version_cache.rs
```diff
@@ -177,8 +177,8 @@ impl<B: BlockT> VersionRange<B> {
 		Self { start: *first.block.header().number(), end: *last.block.header().number(), version }
 	}
 
-	fn contains_block(&self, b: NumberFor<B>) -> bool {
-		(b > self.start && b < self.end) || b == self.start || b == self.end
+	fn contains_block(&self, b: &NumberFor<B>) -> bool {
+		(self.start..=self.end).contains(b)
```
emostov, line 181: Does using the range have space or time complexity overhead relative to boolean logic?

What not use `b >= self.start && b <= self.end`?

emostov (commented):
lgtm modulo a nit I am not sure about concerning `RangeBounds` vs using boolean logic (see comment)

substrate-archive-backend/src/runtime_version_cache.rs
dvdplm: Under the hood they should be exactly the same. 
The idea here is if someone someday figures out a way to optimize `b >= self.start && b <= self.end` we'll get that optimization for free as soon as it lands in the stdlib/compiler. Likely moot in this case, but you never know!

dvdplm: (It also makes clippy STFU)

substrate-archive-backend/src/runtime_version_cache.rs
dvdplm: They are, yes. It's because of how `AsRef` is defined for `Vec<T>` and `[T]` (I think; others can correct me).

PR #171, opened 2021-01-13
Ignore update of some git sources

koushiro:

Close #172
Close #173
Close #174
Close #176

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,14 @@ version: 2
 updates:
   - package-ecosystem: "cargo"
     directory: "/"
+    # Handle updates for crates from github.com/paritytech/substrate manually.
+    ignore:
+      - dependency-name: "substrate-*"
+      - dependency-name: "sc-*"
+      - dependency-name: "sp-*"
+      - dependency-name: "frame-*"
+      - dependency-name: "fork-tree"
+      - dependency-name: "pallet-*"
     schedule:
       interval: "daily"
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,14 +2,16 @@ version: 2
 updates:
   - package-ecosystem: "cargo"
     directory: "/"
-    # Handle updates for crates from github.com/paritytech/substrate manually.
+    # Handle updates for crates from github.com/paritytech/substrate
+    # and github.com/paritytech/polkadot manually.
     ignore:
       - dependency-name: "substrate-*"
       - dependency-name: "sc-*"
       - dependency-name: "sp-*"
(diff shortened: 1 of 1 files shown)


koushiro:
@insipx @dvdplm @niklasad1 PTAL

.github/dependabot.yml
```diff
@@ -2,6 +2,14 @@ version: 2
 updates:
   - package-ecosystem: "cargo"
     directory: "/"
+    # Handle updates for crates from github.com/paritytech/substrate manually.
+    ignore:
+      - dependency-name: "substrate-*"
```
niklasad1, line 7: would nice to ignore to update git dependencies in general I guess but this should do it for now I guess.

dvdplm: I struggle to find the up-to-date docs for dependabot. Best I could find was [this](https://dependabot.com/docs/config-file/#ignored_updates) which uses a different syntax; @koushiro do you know where I should look?

I think the intention is correct here: we prefer to manage the substrate dependencies manually or we'll be approving trivial PRs all day every day. If the `ignore` stuff shown here works well I think we should go with that, trusting that substrate keeps its dependencies up to date (and ourselves to update it regularly).

One alternative route would be to auto-merge all substrate commits (this is what `polkadot` does right?) and live with the breakage.

Would be good to hear from @insipx on this.

koushiro: I refer to the `dependabot.yml` of `polkadot`

insipx: In addition to auto-merging we can `@dependabot ignore this dependency` in the comments which will ignore updating only that dependency. Eventually we'll work through all the substrate deps we don't want. I could see ignoring only dependencies with `substrate-` in it becoming confusing, since libraries like `substrate-subxt` exist which aren't necessarily what we want to ignore. In addition, most substrate dependencies are named like `sp-xx`. `sc-xx` or `frame-xx` anyway, so I don't think this would successfully squash all the PR's.

On auto-merging I kind of prefer doing it manually, it's like a TODO item whenever there's a new Polkadot/Substrate release, and it keeps master stable for-the-most-part. That said i'm ok with the initial slog of using dependabot commands to ignore specific dependencies, if it means knowing exactly which dependencies are ignored

insipx: After further inquiry using @dependabot ignore will ignore a dependency only until it's updated again, so that's actually not what I thought it was

In that case I'd agree with @koushiro that copying Polkadot's dependabot.yml is the best way forward, i'd like to keep master as stable as a master branch can be

insipx (approved):

niklasad1 (approved):

Issue #185, opened 2021-01-14
Make number of Database Actors/Spawned Database Tasks dynamic


insipx:
Database tasks should be based on number of CPU's or maybe even configurable in the config file

https://github.com/paritytech/substrate-archive/blob/4ecb1b4bd02a3351041c9fb5f18d14509e82b6cb/substrate-archive/src/actors.rs#L230

Issue #184, opened 2021-01-14
Merge/Remove ActorPool in favor of more manual management within DatabaseActor


insipx:
In an effort to reduce waiting on database operations, the `ActorPool` module was created in order to manage and spinoff multiple `DatabaseActor` Instances, each with their own mailbox/able to execute database operations. Although it's generic for any actor, the layer of indirection that `ActorPool` introduces is unnecessary and can be simplified. It was made with the thought that it could be used for more than the DatabaseActor, and maybe even offer a layer of generic redundancy. In practice it's used only with the `DatabaseActor`, and didn't prove very useful for much else. Because its tasks are spawned opaquely in the global scope and handles are not kept, pending database futures are difficult to cancel. This is one of the culprits behind archive hanging on shutdown; if there is a long-running task in the background, it cannot/is not canceled. Furthermore, most functions in `ActorPool` sit `#[unused]`. SQLx internally also has some issues surrounding cancelling futures, so this might not be a catch-all solution to the problem.

### **Pros**:
 - Generic, works with any Actor
 - Fairly small, simple to understand
 
### **Cons**:
 - Introduces a layer of indirection that can be confusing when looking at the code
   - syntax like `send(some_type.into()).await?.await?` less than satisfactory
   - fills more message queues than needed (maybe increases memory usage since queues are unbounded)
 - Can't cancel long-running tasks
 
### **Alternative**:

`DatabaseActor` can handle spawning database tasks itself. The amount of tasks spawned  at any one time may be configured, and handles can be kept so that they may be canceled later. This will make syntax surrounding sending to the database neater, and make what's actually happening clearer

Issue #161, opened 2021-01-12
`polkadot-archive`: will stuck with log `Doesn't contain metadata`


koushiro:
I encountered the following log when I run the `polkadot-archive` (eac41861b674cf90b1f7ce8de057b64928ea7851) every time  from scratch. It can be seen from the log that `batch_block_handler` has fallen into an infinite loop. It seems to work normally again when I restart `polkadot-archive`.

## Command
```
./polkadot-archive -c polkadot.toml --chain=polkadot -vvv 
```

## Log
```
[01:34] INFO Doesn't contain metadata
[01:34] INFO Doesn't contain metadata
...
[10:15] INFO Doesn't contain metadata
[10:15] INFO Doesn't contain metadata
```

## batch_block_handler

```rust
	// Returns true if all versions are in database
	// false if versions are missing
	async fn db_contains_metadata(blocks: &[Block<B>], conn: &mut DbConn) -> Result<bool> {
		let specs: hashbrown::HashSet<u32> = blocks.iter().map(|b| b.spec).collect();
		let versions: hashbrown::HashSet<u32> = queries::get_versions(conn).await?.into_iter().collect();
		Ok(specs.is_subset(&versions))
	}

	async fn batch_block_handler(&self, blks: BatchBlock<B>) -> Result<()>
	where
		NumberFor<B>: Into<u32>,
	{
		let mut conn = self.db.conn().await?;
		while !Self::db_contains_metadata(blks.inner(), &mut conn).await? {
			log::info!("Doesn't contain metadata");
			smol::Timer::after(Duration::from_millis(50)).await;
		}
		std::mem::drop(conn);
		self.db.insert(blks).await?;
		Ok(())
	}
```

## PostgreSQL
```sql

polkadot-archive=> select spec,min(block_num),max(block_num),count(*) from blocks 
polkadot-archive=> where spec in (select version from metadata)
polkadot-archive=> group by spec;
 spec |  min   |  max   | count  
------+--------+--------+--------
    0 |      1 |  29230 |  29230
    1 |  29231 | 188835 | 159605
    5 | 188836 | 199404 |  10569
    6 | 199405 | 200000 |    596
(4 rows)

polkadot-archive=> select count(*) from blocks;
 count  
--------
 200000
(1 row)
```

koushiro:
I have added some logs. The above problem seems to be caused by calling `pool.acquire()` to get a connection, which causes a failure to insert data into the database.

```
[...]

PR #186, opened 2021-01-14
Tweaks to Indexing & small bugfixes

insipx:
- Decrease max tasks from 500 to 64. Having 500 tasks queued at once, while decreasing the amount of database queries, is unnecessary/uses more memory than it needs to.

- Increase timeout from 5 to 20 seconds. This is mostly inconsequential, however while running archive I noticed that occasionally some tasks would take 10-15 seconds to complete if it happened upon a long-running database operation. This results in some warnings which occurs less often with a higher timeout.

- Decrease Spawned Database actors from 8 to 4. Not much performance difference between the two (not that I noticed). See #185 and #184  however.

- Don't pollute block indexer with more `Crawl` messages than required
  - rather than `ctx.notify_interval` the blocks actor now spawns a future that awaits the `crawl` so that there is no more than 1 Crawl queued at once. A crawl can take anywhere from 10-335 seconds depending on number of blocks being crawled + inserted at once. In my experience, 10K blocks takes ~20seconds, 100K takes 335 seconds. I looked into trying to optimize postgres inserts on the blocks table for this, but it is a more difficult task than it seems.

- use `handle_while` in storage aggregator
   - afaiu this should handle pushing more storage to the `StorageAggregators` queue while waiting for the current database task (initiated by the `send` to the database) to complete. This keeps the Actors mailbox less full. In my testing it seemed to resulted in more consistent storage indexing.

- adding `awaits` in places where we should await. In `StorageAggregator`, we now wait for the storage to be inserted before sending another database task. In `Metadata` actor we wait for blocks to be inserted before moving on to the next message.
This should decrease the amount of database connections we are using, as well as reduce memory use, since we don't have a bunch of tasks queued for database insertion. May help quell #161

### DIFF
--- a/substrate-archive/src/actors/workers/metadata.rs
+++ b/substrate-archive/src/actors/workers/metadata.rs
@@ -80,7 +80,7 @@ impl<B: BlockT + Unpin> MetadataActor<B> {
 		for b in versions.iter() {
 			self.meta_checker(b.spec, b.inner.block.hash()).await?;
 		}
-		self.addr.send(blks.into()).await?;
+		self.addr.send(blks.into()).await?.await;
 		Ok(())
 	}
 }
--- a/bin/polkadot-archive/Cargo.lock
+++ b/bin/polkadot-archive/Cargo.lock
@@ -4420,7 +4420,7 @@ dependencies = [
  "cfg-if 0.1.10",
  "cloudabi 0.0.3",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.1.57",
  "rustc_version",
  "smallvec 0.6.13",
  "winapi 0.3.9",
@@ -4435,7 +4435,7 @@ dependencies = [
  "cfg-if 0.1.10",
  "cloudabi 0.0.3",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.1.57",
  "smallvec 1.6.1",
  "winapi 0.3.9",
 ]
@@ -4450,7 +4450,7 @@ dependencies = [
  "cloudabi 0.1.0",
  "instant",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.1.57",
  "smallvec 1.6.1",
  "winapi 0.3.9",
 ]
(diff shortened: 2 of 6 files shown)


koushiro:
@insipx  Maybe you could fix the `idle_timeout` config in this PR? It doesn't match the comment.

https://github.com/paritytech/substrate-archive/blob/4ecb1b4bd02a3351041c9fb5f18d14509e82b6cb/substrate-archive/src/database.rs#L59

dvdplm (approved):
LGTM.
If anything I think importing `Duration` to get rid of some `std::time::` line noise. ;)

substrate-archive/src/actors/workers/metadata.rs
```diff
@@ -80,7 +80,7 @@ impl<B: BlockT + Unpin> MetadataActor<B> {
 		for b in versions.iter() {
 			self.meta_checker(b.spec, b.inner.block.hash()).await?;
 		}
-		self.addr.send(blks.into()).await?;
+		self.addr.send(blks.into()).await?.await;
```
niklasad1, line 83: dumb question: is this actix related which is some kind of wrapper type that implements future which return an inner future?

It's really easy miss calling the inner future FWIW.

niklasad1:
> @insipx Maybe you could fix the idle_timeout config in this PR? It doesn't match the comment.

Let's do it in another PR

niklasad1 (approved):

substrate-archive/src/actors/workers/metadata.rs
koushiro: @insipx @niklasad1  the same bug

https://github.com/paritytech/substrate-archive/blob/4ecb1b4bd02a3351041c9fb5f18d14509e82b6cb/substrate-archive/src/actors/workers/storage_aggregator.rs#L95

insipx: Yeah, @niklasad1 It's a module I made to run multiple actors at once. Planning to fix it soon, described in issue #184 though

PR #192, opened 2021-01-19
Improve ArchiveBuilder

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

* improve `ArchiveBuilder`
* remove useless `simple_db.rs` file
* move `Archive` trait into `archive` module

PR #194, opened 2021-01-21
Reorg crates

koushiro:
~~waiting for #192 to be merged firstly.~~

* reorg the crates: 
  * move `ReadOnlyDB` trait into `substrate-archive-backend`
  * move `types`, `models` and `util` modules into `substrate-archive`
  * merge `block_exec` and `tasks`, `sql_block_builder` and `models` 
  * remove `substrate-archive-common`
* ~~improve error handling (#194)~~


insipx:
Why get rid of `common`?

Ideally I would like to merge this after #134 , since it would be a PITA to merge all the reorg changes into another big PR

koushiro:
> Why get rid of `common`?
> 
> Ideally I would like to merge after #134 , since it would be a PITA to merge all the reorg changes into another big PR

IMO, `substrate-archive-common` is a boring crate, it just extracts some **non-public** parts from the two crates of `substrate-archive-backend` and `subtrate-archive` (like `ReadOnlyDB` trait used by backend, `types`, `models` and `logger` used by archive), and also messy error types.

I don't mind one more rebase modification after merging #134

koushiro:
@insipx PTAL

insipx (approved):

insipx:
Overall good changes, condenses code where it was 'hidden' otherwise

koushiro:
@dvdplm @niklasad1 PTAL

substrate-archive/src/wasm_tracing.rs
```diff
@@ -164,6 +164,7 @@ impl TraceHandler {
 
 	// we need this because we don't know the values until after tracing has been executed
 	/// Returns true if a span is part of an enabled Target. Checks WASM in addition to the spans target.
+	#[allow(clippy::suspicious_operation_groupings)]
```
dvdplm, line 167: Is this about silencing the complaint on line 175 (suggestion: "use t.level")? If so that seems to be a clippy bug and not something worth silencing like this imo.

substrate-archive/src/archive.rs
```diff
@@ -29,13 +29,12 @@ use sp_runtime::{
...
 
-use substrate_archive_backend::{runtime_api, ReadOnlyBackend, RuntimeConfig, TArchiveClient};
-use substrate_archive_common::{util, ReadOnlyDB};
+use substrate_archive_backend::{runtime_api, ReadOnlyBackend, ReadOnlyDB, RuntimeConfig, TArchiveClient};
 
 use crate::{
 	actors::{System, SystemConfig},
 	error::Result,
-	migrations,
+	migrations, util,
```
dvdplm, line 37:
```suggestion
	migrations, 
	util,
```

substrate-archive/src/logger.rs
```diff
@@ -1,39 +1,18 @@
-// Copyright 2017-2021 Parity Technologies (UK) Ltd.
```
dvdplm, line 1: All files need license headers. Please revert.

substrate-archive/src/logger.rs
```diff
@@ -1,39 +1,18 @@
...
-
-// substrate-archive is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-
-// You should have received a copy of the GNU General Public License
-// along with substrate-archive.  If not, see <http://www.gnu.org/licenses/>.
-
-//! logging and general utilities
```
dvdplm, line 17: All modules should have module documentation; if anything improve and extend, but please do not remove any existing docs.

dvdplm (changes requested):
Not sure I see the value of doing this. It's @insipx's call but I'm not convinced.

substrate-archive/src/archive.rs
koushiro: it's result of command `cargo fmt`

substrate-archive/src/wasm_tracing.rs
koushiro: Yes, it seems to be a clippy bug (rust-lang/rust-clippy#6722) of rust 1.50, which needs to be fixed by the upstream

koushiro: @dvdplm PTAL

koushiro:
I will improve the running configuration based on this PR, so hope the PR can be reviewed and merged ASAP @insipx @dvdplm

dvdplm (approved):

substrate-archive-backend/src/database.rs
```diff
@@ -23,7 +23,24 @@ use kvdb::KeyValueDB;
...
+
+// Archive specific K/V database reader implementation
+pub trait ReadOnlyDB: Send + Sync {
+	/// Read key/value pairs from the database
+	fn get(&self, col: u32, key: &[u8]) -> Option<Vec<u8>>;
+	/// Iterate over all blocks in the database
+	fn iter<'a>(&'a self, col: u32) -> Box<dyn Iterator<Item = KeyValuePair> + 'a>;
+	/// Catch up with the latest information added to the database
+	fn catch_up_with_primary(&self) -> io::Result<()>;
+	// Open database as read-only
```
niklasad1, line 39:
```suggestion
	/// Open database as read-only
```

substrate-archive/src/error.rs
```diff
@@ -13,18 +13,21 @@
...
-use std::{env, fmt, io};
+use std::{env, fmt, io, num};
 use thiserror::Error;
 
 pub type Result<T, E = ArchiveError> = std::result::Result<T, E>;
 
 /// Substrate Archive Error Enum
 #[derive(Debug, Error)]
 pub enum ArchiveError {
+	// Rust std error
```
niklasad1, line 24: I fine with removing the comment or fixing the comment

```suggestion
	// Rust std io error
```

insipx:
I think this PR is good, and I greatly appreciate your continued contributions.

I think the reason for the delay of this PR is that many of the included changes are stylistic/aesthetic/organizational. While often they make code more readable, substrate-archives codebase is nonetheless under flux and will probably change significantly in the future anyway. So it's sort of sisyphean to try and make the structure perfect now, when there are things like type-decoding/substrate runtime changes/paritydb/pluggable backends in the pipeline that will change the structure anyways. These features also move the project forward toward it's end-goal rather than stagnating on organization. I know it can be annoying to work in a codebase where things aren't where they 'should' be, or where it seems inefficient, but I don't think now is the right time to focus on such changes.


TL;DR I think we should stick with PR's that move the project forward toward it's goal of 'Generic Indexer pluggable for all Substrate Chains' rather than stagnate on organizing the codebase 'correctly'.

In that vein I think PR #227 (I still have to read through all of it, though) is in a better direction, those changes make it clearer to the end user, make config options easier to use and add features which were lacking before (and close two open issues!).


I'll approve this PR but will be more weary towards more stylistic changes in the future

koushiro:
> TL;DR I think we should stick with PR's that move the project forward toward it's goal of 'Generic Indexer pluggable for all Substrate Chains' rather than stagnate on organizing the codebase 'correctly'.

@insipx  Thank you for your detailed reply, I totally agree with it, this will also be what I expected to see

PR #199, opened 2021-01-25
Update Substrate / Deny Dead Code / CI

insipx:
Updates substrate dependencies
  - checking the older version of substrate with `--tests` wouldn't compile because of something in one of the substrate proc macros, so I updated substrate.

- Adds #[deny(dead_code)] to subcrates
- adds cargo checking for tests

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --tests
+          args: --all-targets
 
       - name: Check Code For `node-template-archive`
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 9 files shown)


.github/workflows/rust.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
+          args: --tests
```
dvdplm, line 42: What do you think about going full-monty here and do `--all-targets`? Docs say: "Check all targets. This is equivalent to specifying --lib --bins --tests --benches --examples."

dvdplm (approved):

.github/workflows/rust.yml
insipx: works for me!

niklasad1 (approved):

PR #227, opened 2021-02-22
Improve archive config

### LINKED ISSUE #114: Do not hardcode substrate-archive local directory path
Archive puts it's logs in `~/.local/share/substrate_archive` by default, and doesn't offer any way to specify it's own data directory where it can store rocksdb_secondary/paritydb data or logs. It would be nice to have this as a configurable in the ArchiveBuilder.

Also there are some places where we use the same path but not the same method, EX:
https://github.com/paritytech/substrate-archive/blob/master/src/archive.rs#L165

https://github.com/paritytech/substrate-archive/blob/master/src/util.rs#L48

### LINKED ISSUE #140: Add File Logging CLI Command
Currently there is not option to disable or enable outputting logs to a file for polkadot-archive/node-template-archive. Over time, this log file can grow quite large if it is not being used. There should be a CLI argument, and logging to a file It should be disabled by default and enabled only when it is needed.

Logging to a file includes some extra information about the execution of substrate archive, like line numbers/files loglines came from, more targets enabled, etc. It is mostly useful when debugging.

### LINKED ISSUE #185: Make number of Database Actors/Spawned Database Tasks dynamic

### LINKED ISSUE #209: Make Interpreted vs Compiled WASM a CLI Flag
Currently substrate uses Interpreted as Default, but it would be nice to provide an easier way to switch to compiled.

koushiro:
~~waiting for #194 to be merged firstly.~~

config:
```toml
[chain]
# Must be an absolute path to the folder where polkadot/kusama/and westend chain data is stored
# Can also be specified via the `CHAIN_DATA_DB` environment variable
data_path = "/home/insipx/.local/share/polkadot/chains/polkadot/db"

# How much should the read-only database keep in cache (MB)
# Optional, default: 128
#cache_size = 128

# RocksDB secondary directory, output will be `./output/rocksdb_secondary/`
# Optional, default: /<local>/substrate_archive/rocksdb_secondary/
#rocksdb_secondary_path = "./output/"

#[runtime]
# Specification of different methods of executing the runtime Wasm code.
# Optional, "Interpreted" or "Compiled", default: "Interpreted"
#exec_method = "Interpreted"

# Number of threads to dedicate for executing blocks
# Optional, default: the number of logical system threads
# More BlockWorkers requires that you also increase the number of WASM pages
# else the wasm executor will run out of memory.
# This also increases substrate-archives memory usage.
# Generally, you want 32 pages per block worker
#block_workers = 8

# Number of 64KB Heap Pages to allocate for WASM execution
# Optional, default: 64 * the number of logical system threads
#wasm_pages = 2048

[database]
# Database url.
# Each chain needs it's own PostgreSQL database
# A fully synced Kusama PostgreSQL database (~2.8Million blocks) takes up around 70GB
# These variables can also be exposed with DB_NAME environment variable
# Can also be specified via the `DATABASE_URL` environment variable
# More info on the wiki: https://github.com/paritytech/substrate-archive/wiki/1.)-Requirements
url = "postgres://postgres:123@localhost:5432/polkadot-archive"

#[log]
# Optional log level of stdout, default: "DEBUG"
#std = "DEBUG"

# Optional file log.
#[log.file]
# Optional log level of file, default: "DEBUG"
#level = "DEBUG"
# Optional log file directory path, default: "/<local>/substrate_archive/"
#dir = "./output/"
# Optional log file name, default: "archive.log"
#name = "archive.log"

#[control]
# Number of database actors to be spawned.
# Optional, default: 4
#db_actor_pool_size = 4

# Number of threads to dedicate for executing tasks
# Optional, default: the number of logical system threads
#task_workers = 8

# Timeout to wait for a task to start execution.
# Optional, default: 20 seconds
#task_timeout = 20

# Maximum number of tasks to queue in the threadpool.
# Optional, default: 64
#max_tasks = 64

# Maximium number of blocks to load and insert into database at a time.
# Useful for controlling memory usage.
# Optional, defaults: 100,000
#max_block_load = 100000

#[wasm_tracing]
# Targets for tracing.
#targets = '''wasm_tracing,pallet,frame,state'''

# Folder where tracing-enabled WASM binaries are kept.
#folder = ""
```

Close #114 
Close #140 
Close #185
Close #209


koushiro:
@insipx @niklasad1 @dvdplm PTAL

bin/polkadot-archive/archive.toml
```diff
@@ -0,0 +1,68 @@
+[chain]
+# Must be an absolute path to the folder where polkadot/kusama/and westend chain data is stored
+# Can also be specified via the `CHAIN_DATA_DB` environment variable
+data_path = "/home/insipx/.local/share/polkadot/chains/polkadot/db"
+
+# How much should the read-only database keep in cache (MB)
+# Optional, default: 128
+cache_size = 128
+
+# Rocksdb secondary directory, output will be `./output/rocksdb_secondary/`
```
insipx, line 10:
```suggestion
# RocksDB secondary directory, output will be `./output/rocksdb_secondary/`
```

insipx (commented):
Some nits here and there but the code itself looks good.

I think the big user-facing breaking changes here are specifying the backend directory with an environment variable, CHAIN_DATA_DB and the postgres db with DATABASE_URL rather than specifying some switches in the config, ie `polkadot-db`, `kusama-db`, `westend-db`, and the _root_ path of the RocksDB directory.

I was never sure which was more user friendly, however I think it saves alot of time typing out a database url and rocksdb database if you want to index multiple chains. Environment variables seem more standard and are often used in conjunction with docker though. Thoughts on this, IE.. Worth making people re-work configs? @dvdplm . I don't think it's too big a deal anyway though.

Other than that, I like seeing the reduction in code with structopt and log directory handling/consolidation of arguments was long-overdue

koushiro:
> Some nits here and there but the code itself looks good.
> 
> I think the big user-facing breaking changes here are specifying the backend directory with an environment variable, CHAIN_DATA_DB and the postgres db with DATABASE_URL rather than specifying some switches in the config

Actually, the current changes are still compatible with the situation where the configuration file is not specified and only the environment variable `CHAIN_DATA_DB` and `DATABASE_URL` is set

substrate-archive/src/archive.rs
```diff
@@ -73,116 +128,192 @@ where
...
+			Self::default()
+		}
+	}
+
+	/// Specify a chain spec name and id for storing metadata about the running archiver
+	/// in a persistent directory.
 	///
 	/// # Default
-	/// Defaults to storing metadata in a temporary directory.
+	/// defaults to storing metadata in a temporary directory.
```
dvdplm, line 159:
```suggestion
	/// Defaults to storing metadata in a temporary directory.
```

substrate-archive/src/archive.rs
```diff
@@ -73,116 +128,192 @@ where
...
-		self.chain_data_path = path.map(Into::into);
+	pub fn chain_data_path<S: Into<PathBuf>>(mut self, path: S) -> Self {
+		self.config.chain.data_path = Some(path.into());
+		self
+	}
+
+	/// Set the amount of cache RocksDB should keep.
+	///
+	/// # Default
+	/// defaults to 128MB
```
dvdplm, line 177:
```suggestion
	/// Defaults to 128MB.
```

(etc…)

substrate-archive/src/archive.rs
```diff
@@ -73,116 +128,192 @@ where
...
+	pub fn cache_size(mut self, cache_size: usize) -> Self {
+		self.config.chain.cache_size = cache_size;
+		self
+	}
+
+	/// Set the path of RocksDB secondary directory according to the chain spec.
```
dvdplm, line 183:
```suggestion
	/// Set the path to the secondary RocksDB database directory according to the chain spec.
```

"according to the chain spec": not sure what is meant by this. Relative to the chain spec?

substrate-archive/src/archive.rs
```diff
@@ -73,116 +128,192 @@ where
...
 	/// # Default
 	/// defaults to value of the environment variable DATABASE_URL
-	pub fn pg_url<S: Into<String>>(mut self, url: Option<S>) -> Self {
-		self.pg_url = url.map(Into::into);
+	pub fn pg_url<S: Into<String>>(mut self, url: S) -> Self {
+		self.config.database = Some(DatabaseConfig { url: url.into() });
 		self
 	}
 
-	/// Set the amount of cache Rocksdb should keep.
```
dvdplm, line 146: Maybe I missed it, but where do we set the cache size now?

substrate-archive/src/archive.rs
```diff
@@ -247,34 +356,46 @@ where
...
+		self.config.runtime.wasm_runtime_overrides = self.config.wasm_tracing.as_ref().and_then(|c| c.folder.clone());
 
-		let client = Arc::new(runtime_api::<B, R, D, DB>(db.clone(), config)?);
+		// configure substrate client and backend
+		let client = Arc::new(runtime_api::<B, R, D, DB>(db.clone(), self.config.runtime)?);
 		let backend = Arc::new(ReadOnlyBackend::new(db, true));
 		Self::startup_info(&*client, &*backend)?;
 
+		// config postgres database
+		const POSTGRES_VAR: &str = "DATABASE_URL";
```
dvdplm, line 384:
```suggestion
		const DATABASE_URL: &str = "DATABASE_URL";
```

substrate-archive/src/archive.rs
```diff
@@ -247,34 +356,46 @@ where
...
-		let chain_path = self.chain_data_path.unwrap_or_else(|| {
-			env::var(CHAIN_DATA_VAR).expect("CHAIN_DATA_DB must be set if not passed initially.").into()
-		});
+	pub fn build(mut self) -> Result<impl Archive<B, DB>> {
+		// config logger
+		logger::init(self.config.log.clone())?;
+		log::debug!("Archive Config: {:?}", self.config);
+
+		// config chain
+		const CHAIN_DATA_VAR: &str = "CHAIN_DATA_DB";
```
dvdplm, line 365:
```suggestion
		const CHAIN_DATA_DB: &str = "CHAIN_DATA_DB";
```

substrate-archive/src/archive.rs
```diff
@@ -304,3 +427,32 @@ where
 		Ok(())
 	}
 }
+
+/// Create rocksdb secondary directory if it doesn't exist yet.
```
dvdplm, line 431:
```suggestion
/// Create the secondary rocksdb directory if it doesn't exist yet.
```

substrate-archive/src/archive.rs
```diff
@@ -304,3 +427,32 @@ where
...
+/// If the ChainSpec is not specified, a temporary directory is used.
+/// Return path to that directory
```
dvdplm, line 433:
```suggestion
/// Returns the path to that directory
```

substrate-archive/src/archive.rs
```diff
@@ -304,3 +427,32 @@ where
...
+///
+/// # Panics
+///
+/// Panics if the directories fail to be created.
```
dvdplm, line 437:
```suggestion
/// Panics if the directory creation fails.
```

substrate-archive/src/archive.rs
```diff
@@ -304,3 +427,32 @@ where
...
+			Ok(db_path)
+		}
+		(None, Some(spec)) => {
+			let mut path = substrate_archive_default_dir();
+			path.extend(&["rocksdb_secondary", &spec.name, &spec.id]);
+			fs::create_dir_all(path.as_path())?;
+			Ok(path)
+		}
+		_ => {
+			// TODO: make sure this is cleaned up on kill
```
dvdplm, line 452: Please add issue (or remove/fix the todo).

substrate-archive/src/logger.rs
```diff
@@ -16,26 +16,59 @@
...
+		Self { level: default_file_log_level(), dir: None, name: default_file_log_name() }
+	}
+}
+
+fn default_file_log_level() -> log::LevelFilter {
+	log::LevelFilter::Debug
+}
+
+fn default_file_log_name() -> String {
+	"archive.log".into()
```
dvdplm, line 58:
```suggestion
	"substrate-archive.log".into()
```

substrate-archive/src/logger.rs
```diff
@@ -51,39 +84,39 @@ pub fn init(std: log::LevelFilter, file: log::LevelFilter) -> io::Result<()> {
...
+			.level(file.level)
+			.level_for("substrate_archive", file.level)
+			.level_for("cranelift_wasm", log::LevelFilter::Error)
+			.level_for("sqlx", log::LevelFilter::Warn)
+			.level_for("staking", log::LevelFilter::Warn)
+			.level_for("cranelift_codegen", log::LevelFilter::Warn)
+			.level_for("frame_executive", log::LevelFilter::Error)
+			// .level_for("desub_core", log::LevelFilter::Debug)
+			// .level_for("kvdb_rocksdb", log::LevelFilter::Debug)
+			// .level_for("kvdb_rocksdb", log::LevelFilter::Debug)
```
dvdplm, line 104:
```suggestion
```

dvdplm (changes requested):
Looks like an improvement.

Left questions and nitpicks.

substrate-archive/src/archive.rs
koushiro: the `cache_size` method of `ArchiveBuilder`

substrate-archive/src/archive.rs
koushiro: if you specify the path `./substrate-archive/rocksdb_secondary` and chain spec,
the actual path will be `./substrate-archive/rocksdb_secondary/<chain-spec-name>/<chain-spec-id>`.
I will add this comment.

examples/simple.rs
```diff
@@ -4,7 +4,7 @@ use polkadot_service::{kusama_runtime::RuntimeApi, Block, KusamaExecutor};
 use substrate_archive::{Archive, ArchiveBuilder};
 
 pub fn main() {
-    substrate_archive::init_logger(log::LevelFilter::Info, log::LevelFilter::Info);
+    substrate_archive::logger::init(Default::default());
```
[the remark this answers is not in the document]
koushiro: Actually, the default level of logger is `Debug`.

bin/node-template-archive/src/cli_opts.rs
```diff
@@ -14,39 +14,45 @@
...
+use structopt::StructOpt;
 
-#[derive(Clone)]
+use substrate_archive::ArchiveConfig;
+
+#[derive(Clone, StructOpt)]
+#[structopt(author, about)]
 pub struct CliOpts {
-	pub file: Option<PathBuf>,
-	pub log_level: log::LevelFilter,
```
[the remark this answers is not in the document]
koushiro: Yeah, we need to set the log level in the config file here if we want to change the logging level

substrate-archive-backend/src/frontend.rs
```diff
@@ -39,13 +40,63 @@ pub type TArchiveClient<TBl, TRtApi, TExecDisp, D> = Client<TFullCallExecutor<TB
...
+}
+
+// the number of logical cpus in the system
+fn default_block_workers() -> usize {
+	num_cpus::get()
+}
+
+// 64 * (number of logic cpu's)
+fn default_wasm_pages() -> u64 {
+	64 * num_cpus::get() as u64
```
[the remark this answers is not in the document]
koushiro: So you mean that if `wasm_pages` is not configured, just use `64 * block_workers` for it?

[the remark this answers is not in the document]
koushiro: Or use the default configuration of `WasmExecutor` directly if we don't specify the `wasm_pages`?

`sc-client/exector/src/native_exector.rs`:
```rust
/// Default num of pages for the heap
const DEFAULT_HEAP_PAGES: u64 = 1024;

impl WasmExecutor {
	/// Create new instance.
	///
	/// # Parameters
	///
	/// `method` - Method used to execute Wasm code.
	///
	/// `default_heap_pages` - Number of 64KB pages to allocate for Wasm execution.
	/// 	Defaults to `DEFAULT_HEAP_PAGES` if `None` is provided.
	pub fn new(
		method: WasmExecutionMethod,
		default_heap_pages: Option<u64>,
		host_functions: Vec<&'static dyn Function>,
		max_runtime_instances: usize,
	) -> Self {
		WasmExecutor {
			method,
			default_heap_pages: default_heap_pages.unwrap_or(DEFAULT_HEAP_PAGES),
			host_functions: Arc::new(host_functions),
			cache: Arc::new(RuntimeCache::new(max_runtime_instances)),
			max_runtime_instances,
		}
	}
}
```

bin/node-template-archive/archive.toml
```diff
@@ -0,0 +1,73 @@
...
+# Optional, default: /<local>/substrate_archive/rocksdb_secondary/
+rocksdb_secondary_path = "./output/"
+
+[runtime]
+# Specification of different methods of executing the runtime Wasm code.
+# Optional, "Interpreted" or "Compiled", default: "Interpreted"
+#exec_method = "Interpreted"
+
+# Number of threads to dedicate for executing blocks
+# Optional, default: the number of logical system threads
```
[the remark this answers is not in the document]
koushiro: I'm not sure about this, just keep the original configuration.

examples/simple.rs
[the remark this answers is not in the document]
dvdplm: Good point. So ideally we should set `Info` for the main app and I'd prefer the example to explicitly set the level to `Debug`.

substrate-archive-backend/src/frontend.rs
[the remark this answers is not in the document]
dvdplm: > Or use the default configuration of `WasmExecutor` directly if we don't specify the `wasm_pages`?

That is a good idea, let's do that.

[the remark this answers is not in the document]
dvdplm: > So you mean that if `wasm_pages` is not configured, just use `64 * block_workers` for it?

Yes, that's what I had in mind, but I think your suggestion to fall back to the executor defaults is a better rule.

dvdplm (approved):
LGTM, modulo sorting out the default wasm pages.

insipx (approved):

PR #195, opened 2021-01-22
Fix missing genesis block

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

Fix #187

Issue #187, opened 2021-01-15
missing the genesis block


koushiro:
Every time I run `polkadot-archive` from scratch, the database will miss the genesis block.

```sql
polkadot-archive=> select spec,min(block_num),max(block_num),count(*),max(block_num)-min(block_num)+1-count(*) as missing from blocks
polkadot-archive=> where spec in (select version from metadata)
polkadot-archive=> group by spec;
 spec |  min  |  max  | count | missing 
------+-------+-------+-------+---------
    0 |     1 | 29230 | 29230 |       0
    1 | 29231 | 65000 | 35770 |       0
(2 rows)
```

 but when I restart `polkadot-archive`, a missing block will be detected and inserted into the database.

```
[13:57] DEBUG Registered a new runtime version: RuntimeVersion { spec_name: RuntimeString::Owned("polkadot"), impl_nam
e: RuntimeString::Owned("parity-polkadot"), authoring_version: 0, spec_version: 0, impl_version: 0, apis: [([223, 106,
 203, 104, 153, 7, 96, 155], 3), ([55, 227, 151, 252, 124, 145, 245, 228], 1), ([64, 254, 58, 212, 1, 248, 149, 154], 
4), ([210, 188, 152, 151, 238, 208, 143, 21], 2), ([247, 139, 39, 139, 229, 63, 69, 76], 2), ([175, 44, 2, 151, 162, 6
2, 109, 61], 3), ([237, 153, 197, 172, 178, 94, 237, 245], 2), ([203, 202, 37, 227, 159, 20, 35, 135], 2), ([104, 122,
 212, 74, 211, 127, 3, 194], 1), ([171, 60, 5, 114, 41, 31, 235, 139], 1), ([188, 157, 137, 144, 79, 91, 146, 63], 1),
 ([55, 200, 187, 19, 80, 169, 162, 168], 1)], transaction_version: 0 }
[13:57] DEBUG took 4.107681ms to insert 1 blocks
[13:57] INFO 1 missing blocks
[13:57] INFO Took 3.979677024s to load 50000 blocks
[13:57] DEBUG Adding new runtime code hash to cache: 0x3ADB391C9FB52BA7
[13:57] DEBUG Registered a new runtime version: RuntimeVersion { spec_name: RuntimeString::Owned("polkadot"), impl_nam
e: RuntimeString::Owned("parity-polkadot"), authoring_version: 0, spec_version: 26, impl_version: 0, apis: [([223, 106, 
[...]

PR #243, opened 2021-03-11
Improve genesis block

koushiro:
related issue: #187 
previous solution: #195

### DIFF
--- a/substrate-archive/src/actors/workers/blocks.rs
+++ b/substrate-archive/src/actors/workers/blocks.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use xtra::prelude::*;
 
 use sp_runtime::{
-	generic::SignedBlock,
+	generic::{BlockId, SignedBlock},
 	traits::{Block as BlockT, Header as _, NumberFor},
 };
 use substrate_archive_backend::{ReadOnlyBackend, ReadOnlyDB, RuntimeVersionCache};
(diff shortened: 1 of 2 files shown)


substrate-archive/src/actors/workers/blocks.rs
```diff
@@ -74,6 +74,29 @@ where
...
+		} else {
+			Ok(None)
+		}
+	}
+
+	async fn fetch_and_send_genesis_block(&self) -> Result<()> {
+		if let Some(genesis_block) = self.fetch_genesis_block().await? {
+			self.meta.send(genesis_block).await?;
+		} else {
+			log::error!("Couldn't fetch the genesis block from backend");
```
dvdplm, line 95: dq: why isn't this a hard error? I.e. shouldn't we return an error here?

substrate-archive/src/actors/workers/blocks.rs
```diff
@@ -74,6 +74,29 @@ where
...
+	where
+		NumberFor<B>: From<u32>,
+	{
+		let backend = self.backend.clone();
+		if let Some(block) = backend.block(&BlockId::Number(0.into())) {
+			let cache = self.rt_cache.clone();
+			let spec = cache.get(block.block.hash())?.map(|ver| ver.spec_version).unwrap_or_default();
+			Ok(Some(Block { inner: block, spec }))
+		} else {
+			Ok(None)
```
dvdplm, line 87: Why not return an error here?

PR #252, opened 2021-04-08
[WIP] Remove actorpool

### LINKED ISSUE #184: Merge/Remove ActorPool in favor of more manual management within DatabaseActor

insipx:
Removing actorpool to let the Database actor manually handle it's futures that are inserting tasks. This should remove a layer of indirection and help speed up the database, which will be particularly helpful when adding another actor for decoding everything inside of a block + storage.

This PR needs a bit of cleanup since I started working on #253 at around the same time as this PR, so there is some leftover parts. I also haven't fully tested these changes yet (which is the rabbithole that led me down to #253 in the first place)

fixes #184 

I'd like to get these benchmarks/changes in before starting full-steam work on desub, since I think they will help performance significantly in the long-run with the extra load that decoding will add/help us identify regressions in storage execution.


insipx:
Closing this in favor of #254

PR #253, opened 2021-04-08
Benchmarks for Block Execution + Threadpooled block execution

insipx:
I think this is some part of the code that experiences the most fluctuations between releases/pushes, where in one release there may be blocks executing at X/second and the next it decreases. I hope these benchmarks will help us identify whether there is a regression in a PR or not.

There are still some that need to be added, and it's a bit annoying to make parts of the API public where they previously weren't (since criterion cannot bench withing `src/` but I believe it to be worth it).

All benchmarks run from spec 26, where a CSV file is included that has 10,000 blocks starting from block 3,000,000 on Polkadot


insipx:
Closing this, might possibly revisit at a later date

PR #254, opened 2021-04-09
Remove Actorpool

### LINKED ISSUE #184: Merge/Remove ActorPool in favor of more manual management within DatabaseActor

insipx:
Removing actorpool to let the Database actor manually handle it's futures that are inserting tasks. This should remove a layer of indirection and help speed up the database, which will be particularly helpful when adding another actor for decoding everything inside of a block + storage.

This PR needs a bit of cleanup since I started working on #253 at around the same time as this PR, so there is some leftover parts. I also haven't fully tested these changes yet (which is the rabbithole that led me down to #253 in the first place)

fixes #184

I'd like to get these benchmarks/changes in before starting full-steam work on desub, since I think they will help performance significantly in the long-run with the extra load that decoding will add/help us identify regressions in storage execution.

Still needs some work to fix the custom smol executor


insipx:
While I consider this PR ready for review i'll be splitting it up into mutliple PR's in the morning. Leaving it up for now, though.

This PR does a few things, namely get archive into a place where I think it's ready to bear even more inserts at once into postgres that would come with type-decoding integration. It localizes actor control in one place, makes shutting down more consistent (still issues with canceling futures, though), removes indirection and relieves memory use at some critical locations.

- Removes the 'ActorPool' (fixes #184)
- Gets rid of some technical debt:
	- Removes type generics where not needed (mostly on struct definitions)
	- Uses threaded executor for `smol` 
- adds tests for a new query
	- new query is more efficient at finding missing storage changes, and allows for pagination of results to reduce memory usage
	- new SQL function 'HEX_TO_INT' converts between Hexadecimal ASCII and Decimal Representation
	- `_background_tasks` now stores data in `json` rather than compact `rmp` representation. This uses more disk space but 
	allows for querying into the JSON rather than needing to load the blocks and decode them.
	- adds 10,000 blocks from Kusama in CSV format to facilitate testing
	- updates the table function to index on `block_num` rather than `ID`

substrate-archive/src/actors.rs
```diff
@@ -157,15 +144,79 @@ where
...
+		tasks.push(conf.executor.spawn(fut));
+		let blocks = workers::BlocksIndexer::new(conf, db.clone(), metadata.clone()).create(None);
+		let (blocks, fut) = blocks.run();
+		tasks.push(conf.executor.spawn(fut));
+
+		Ok(Actors { storage, blocks, metadata, db, tasks })
+	}
+
+	// Run a future that sends actors a signal to progress every X seconds
+	async fn tick_interval(&self, executor: Arc<smol::Executor<'_>>) -> Result<()> {
```
dvdplm, line 193: Not sure I see where the `x seconds` part is happening, can you point it out for me?

dvdplm (approved):

substrate-archive/src/database/queries.rs
```diff
@@ -178,22 +152,137 @@ pub(crate) async fn get_versions(conn: &mut PgConnection) -> Result<Vec<u32>> {
...
+	nums: &'a [u32],
+	limit: usize,
+) -> impl Stream<Item = Result<Vec<BlockModel>>> + 'a {
+	let nums: Vec<i32> = nums.iter().filter_map(|n| i32::try_from(*n).ok()).collect();
+
+	Box::pin(try_stream! {
+		for page in nums.chunks(limit) {
+			let mut query = String::from("SELECT * FROM blocks WHERE block_num IN (");
+			for (i, num) in page.iter().enumerate() {
+				itoa::fmt(&mut query, *num)?;
```
niklasad1, line 184: are you using itoa for performance reasons here instead of `format!()`?

niklasad1:
> Removes type generics where not needed (mostly on struct definitions)

:+1: ,also quite nice with removing the `BlockT` from the actorpool

substrate-archive/src/actors.rs
```diff
@@ -157,15 +144,79 @@ where
 	}
 }
 
-struct Actors<B: BlockT + Unpin, D: ReadOnlyDb + 'static>
+struct Actors<Block: Send + Sync + 'static, H: Send + Sync + 'static, Db: Send + Sync + 'static> {
```
niklasad1, line 147: are these trait bounds needed?

substrate-archive/src/database/queries.rs
insipx: Yes! This needs to format alotof block numbers

substrate-archive/src/actors.rs
insipx: Yeah, `Address` requires Sync + Send in `xtra`

niklasad1 (approved):
I skimmed it quite fast but looks good

PR #265, opened 2021-04-24
Use a specific release branch instead of the master branch

koushiro:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1774,7 +1774,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "3.0.0"
-source = "git+https://github.com/paritytech/substrate?branch=master#3efe87daba09be18b3596e075c3733fe47bafdd5"
+source = "git+https://github.com/paritytech/substrate?branch=polkadot-v0.8.30#3655f9bb44d75ff205737023760198733a237685"
 dependencies = [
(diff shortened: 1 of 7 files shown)


dvdplm:
I think that `master` should use the latest master branch, but when we cut `substrate-archive` releases we should build with a specific (released) version of polkadot. I am aware of the quick development pace on `polkadot` can make this painful, but still think it's best to make development on `archive` stay as close as possible to the upstream cutting edge. Am I wrong?

@insipx wdyt?

insipx:
Yeah, I think we tried attaching master to specific releases of polkadot in the past, and it caused more headache than it was worth (updating/compile errors/etc). However those issues were caused because the version of substrate pointed at by polkadot dependencies was `master` but I see that has changed now. 

I think for ease & speed of development archive should still point at substrate master. This comes in particular use when we are developing joint features (state-tracing comes to mind) that are being merged into substrate. Currently we would not have to wait for a polkadot release to push to master to get those new substrate features.


i think for substrate releases freezing polkadot versions is what we should switch to doing. We would have to make a release coinciding with each polkadot release, but that would make the compiles/releases generally more consistent/stable

dvdplm:
> i think for substrate releases freezing polkadot versions is what we should switch to doing. We would have to make a release coinciding with each polkadot release, but that would make the compiles/releases generally more consistent/stable

Do you think we can release a `0.5.1` that is pinned to a polkadot release?

dvdplm:
Closing in favour of #267 – thank you @koushiro for the suggestion and the code changes.

PR #271, opened 2021-05-11
Rework tracing, Tests

insipx:
This updates tracing for the latest work in state-tracing PR that went into substrate. To maintain sanity that state-tracing is actually working, I also decided to create a testing system for this ( I think archive needs alot more tests in general).


dvdplm:
What was the reason for closing this?

PR #273, opened 2021-05-11
Refactor Tracing

insipx:
This PR starts the general framework for tests within archive, adding a new crate `test-common` that includes utilities to accomplish this. It's in a separate crate to enable benchmarks with criterion coming with #253, however the functions it exports are useful to test state-tracing as well.

Block execution was refactored to branch off into a different function, `execute_with_tracing` if tracing is enabled, and `into_storage_changes` if not. Previously, we would actually collect some superfluous tracing logs even with tracing disabled since we still ran the tracing dispatcher. This ensures that the dispatcher is only run if tracing is enabled, and coincidentally also makes testing a bit easier.

There are no actual tests included in this PR (I accidentally corrupted my polkadot database and need to resync), but I thought it big enough to merge in the changes for tests, and make a different PR for the actual state-tracing tests.

The downside to this testing method is that the tests will require a synced --pruning=archive polkadot node, since block execution requires read access to the backend database to retrieve storage values, so it might be tricky to integrate with Github Actions. However, I think that's much better than no testing at all

Needs to be merged after #284


dvdplm:
> The downside to this testing method is that the tests will require a synced --pruning=archive polkadot node, since block execution requires read access to the backend database to retrieve storage values,

Why isn't it enough to have a node with _some_ data in it? Why is it critical that the node must be fully synced?

insipx:
> > The downside to this testing method is that the tests will require a synced --pruning=archive polkadot node, since block execution requires read access to the backend database to retrieve storage values,
> 
> Why isn't it enough to have a node with _some_ data in it? Why is it critical that the node must be fully synced?

In this case if we want to include a test for every runtime version v25+ (where tracing-enabled wasm blobs begin) we'd need at least one block representing each runtime version up to the current one. So I guess technically you don't need to be _fully_ synced for _all_ the tests, but as we add tests for new runtime versions it's easier to just say 'you should be fully synced' then remember to update a block number in a comment somewhere

These are also more like integration tests than unit tests, so maybe we should create yet another crate for these rather than have them in `tasks`. _Actually_ unit testing execute_block would require building out a MockRuntimeApi and stuff, but it wouldn't need a synced node

dvdplm (commented):
Overall lgtm. Some questions and usual nitpicks. :)

test-common/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+type PolkadotBackend = Arc<ReadOnlyBackend<Block, SecondaryRocksDb>>;
+
+pub fn get_dot_runtime_api(
+	block_workers: usize,
+	wasm_pages: u64,
+) -> Result<(PolkadotClient, PolkadotBackend), Error> {
+	let chain_path = std::env::var("CHAIN_DATA_DB").expect("must have database as variable CHAIN_DATA_DB");
+	let mut tmp_path = tempfile::tempdir()?.into_path();
+	tmp_path.push("rocksdb_secondary");
+	let db = Arc::new(SecondaryRocksDb::open_database(chain_path.as_str(), 128, tmp_path).unwrap());
```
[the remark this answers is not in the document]
insipx: This is actually a config parameter configurable via the `.toml`, but since this is a test it's just hardcoded. However, even in production I haven't found much benefit to increasing beyond 128

substrate-archive/src/tasks.rs
```diff
@@ -166,9 +186,52 @@ where
...
+		let changes =
+			api.into_storage_changes(&state, None, parent_hash).map_err(ArchiveError::ConvertStorageChanges)?;
+
+		let changes = BlockChanges {
+			storage_changes: changes.main_storage_changes,
+			child_storage: changes.child_storage_changes,
+			block_hash: hash,
+			block_num: number,
+		};
+		// We destroy the Arc and transform the Mutex here in order to avoid additional allocation.
```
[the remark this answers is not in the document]
insipx: Yeah, we would need to drain(..).collect() the contents within the vectors of the mutex, causing an allocation into a different vector (see newer implementation)

substrate-archive/src/tasks.rs
```diff
@@ -166,9 +186,52 @@ where
...
+		// The Arc is cloned into the thread-local tracing subscriber in the scope of `storage`, creating
```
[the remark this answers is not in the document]
insipx: Ah yeah, sorry this is confusing because previously the tracing was collected inside a 

```rust
let storage = {
    // ... block execution code
    return storage_changes;
};
```
but this change removes that. in this case the scope would be this part of the code (it is no longer named)

```rust
{
			let handler = TraceHandler::new(&targets, number.into(), hash.as_ref().to_vec(), span_events.clone());
			let dispatch = Dispatch::new(handler);

			let dispatcher_span = tracing::debug_span!(
				target: "state_tracing",
				"execute_block",
				extrinsics_len = block.extrinsics().len()
			);
			let _guard = dispatcher_span.enter();
			dispatcher::with_default(&dispatch, || {
				let span = tracing::info_span!("block_trace", "trace_block");
				let _enter = span.enter();
				api.execute_block(&id, block)
			})?;
		}

```
We need it to close the `execute_block`span properly and not include anything beyond the block execution


This comment is removed since we no longer unwrap the mutex, though

insipx:
I changed the trace-collection in `execute_with_tracing` to not execute with mutex but instead collect from a `drain(..)`. I was thinking of refactoring this anyway (this is how the collection was originally implemented in archive). I think this way with destroying the Mutex & stuff is not worth the possibility of things to fail, and if it does fail because of some code changes it could be harder to debug.

substrate-archive/src/wasm_tracing.rs
```diff
@@ -73,7 +73,7 @@ pub struct SpanMessage {
 }
 
 /// Finished Trace Data Format. Ready for insertion into a relational database.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct Traces {
 	block_num: u32,
 	hash: Vec<u8>,
```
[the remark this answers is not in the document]
insipx: Yup

substrate-archive/src/tasks.rs
```diff
@@ -221,12 +221,12 @@ where
...
-		// We destroy the Arc and transform the Mutex here in order to avoid additional allocation.
-		// The Arc is cloned into the thread-local tracing subscriber in the scope of `storage`, creating
-		// 2 strong references. When block execution finishes, storage is collected and that reference is dropped.
-		// This allows us to unwrap it here. QED.
-		let traces = Arc::try_unwrap(span_events).map_err(|_| TracingError::NoTraceAccess)?.into_inner();
-		let traces = Traces::new(number.into(), hash.as_ref().to_vec(), traces.events, traces.spans);
+
+		let mut traces = span_events.lock();
+		let spans = traces.spans.drain(..).collect::<Vec<SpanMessage>>();
+		let events = traces.events.drain(..).collect::<Vec<EventMessage>>();
```
dvdplm, line 227: This looks similar to how [we did it in substrate](https://github.com/paritytech/substrate/blob/20f1e369edb860a8287d79dd85a27b50e391b07c/client/tracing/src/block/mod.rs#L253-L268).

insipx: Yeah, in general this implementation brings trace_block stuff in archive closer to what the code looks like in substrate :P

insipx:
Removed alot of the 'live' data that i pulled directly from Polkadot/Kusama chains, since I think that is better suited for Benchmarking/Integration tests. Instead, I made a more direct unit test that doesn't depend on any kind of backend and will run well within Github CI.

substrate-archive/src/tasks.rs
```diff
@@ -141,22 +151,31 @@ where
...
 		// so digest isn't very important (we don't currently index digest items anyway)
 		// popping a digest item has no effect on storage changes afaik
-		let (mut header, ext) = self.block.deconstruct();
+		let (mut header, ext) = block.deconstruct();
 		header.digest_mut().pop();
-		let block = Block::new(header, ext);
+		Ok(BlockPrep { block: Block::new(header, ext), state, hash, parent_hash, number })
+	}
+
+	fn block_into_storage(self) -> Result<BlockChanges<Block>, ArchiveError> {
```
emostov, line 176:
```suggestion
	fn into_storage(self) -> Result<BlockChanges<Block>, ArchiveError> {
```
I don't think it really matters, but I think it would be nice to have the fn name start with `into_*`

substrate-archive/src/tasks.rs
```diff
@@ -87,8 +87,8 @@ pub struct BlockChanges<Block: BlockT> {
 	/// In memory arrays of storage values for multiple child tries.
 	pub child_storage: ChildStorageCollection,
 	/// Hash of the block these changes come from
-	pub block_hash: Block::Hash,
-	pub block_num: NumberFor<Block>,
+	pub hash: Block::Hash,
+	pub number: NumberFor<Block>,
```
emostov, line 91:
```suggestion
    /// Number of the block these changes come from
	pub number: NumberFor<Block>,
```

substrate-archive/src/tasks.rs
emostov: Actually, given that we have `execute_with_tracing` that does the same thing but with tracing (correct me if I'm wrong), it would nice if the names paired. So either `into_storage` & `into_storage_with_traces` or `execute` & `execute_with_tracing` etc

substrate-archive/src/wasm_tracing.rs
```diff
@@ -197,6 +199,20 @@ impl TraceHandler {
 		self.span_events.lock().spans.push(span);
 		Ok(())
 	}
+
+	/// Start tracing with the predicate `fun`.
+	/// Consumes this TraceHandler.
+	pub fn scoped_trace<T>(self, fun: impl FnOnce() -> Result<T>) -> Result<(Vec<SpanMessage>, Vec<EventMessage>, T)> {
+		let span_events = self.span_events.clone();
+		let dispatch = Dispatch::new(self);
+		let res = dispatcher::with_default(&dispatch, fun)?;
```
emostov, line 208: dq: what is `res` here?

substrate-archive/src/wasm_tracing.rs
```diff
@@ -332,3 +348,58 @@ fn parse_target(s: &str) -> (String, Level) {
...
+					&mut ext,
+					true,
+					"test_trace_handler",
+					&[],
+				)
+				.unwrap();
+			Ok(())
+		})?;
+		assert!({
+			spans[0].name == "im_a_span"
```
emostov, line 392: dq: where are we getting the dummy data from? ext?

insipx: the `lib.rs` of this test-wasm crate: https://github.com/paritytech/substrate-archive/tree/insipx/refactor-tracing/test-common/test-wasm just compiles to a wasm blob. That blob is available under a constant and is executed with the default substrate `WasmExecutor` further above the code, and surrounded with the Trace subscriber

emostov: I think I see, its from `test_trace_handler`?

emostov (approved):
It looks good to me overall. Feel free to ping me for review again tho

substrate-archive/src/wasm_tracing.rs
insipx: `dispatcher::with_default` accepts a closure which returns `T`, and in case that is useful further up the chain of function calls we catch it and return it here as `res`(short for result, but maybe a different name would be better)

substrate-archive/src/wasm_tracing.rs
insipx: yup!

substrate-archive/src/wasm_tracing.rs
emostov: Ok, good to know. For the record I think its clear that `res` => result.

insipx:
> It looks good to me overall, just a few questions for my understanding. Feel free to ping me for review again though.
> 
> I like the idea of having "live data" come in a separate PR.

Yeah, same, and  think it will be necessary as archive grows. I think it was becoming a bit too complicated and not thought out all the way through for this PR though.

insipx:
This PR needs to be merged after #284, since I pulled those changes in

insipx:
Once CI passes this is ready for merge too.

Removed `CurrentSpan` and updated tracing-subscriber to the latest

dvdplm (approved):
Good stuff here. I really like how you sorted out the `current_span` stuff in https://github.com/paritytech/substrate-archive/pull/273/commits/65c632474953502c82b47f3d2d260ccba283cdfe

PR #279, opened 2021-05-20
Make releases EZ

insipx:
PR adds a github action workflow which automates releases. It will run whenever the branch `release` gets pushed to. Once it's pushed to, the workflow:

- Builds `polkadot-archive` in `--release`, packages binary in `.deb`and also standalone linux (as with all previous releases)
- Reads from Changelog to fill in release notes in body of release, as well as assign release tag & name

This PR also fixes CHANGELOG to follow keep a changelog, and creates a separate CHANGELOG for polkadot-archive. This will make it easier to update in the future I think.

This won't publish the release, but it will create a draft that we can look over and modify and then publish from Github UI. This is my preferred way, just because it forces double-checking things, but I'm open to other opinions

I suspect  this could further be automated by adding a `diener`step to automatically change all substrate/polkadot refs to the latest release with some more github-action magic. 

I was  testing this stuff out earlier in a personal repo before creating the action (Release branch): https://github.com/insipx/test-gh-action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,8 @@
 name: Release CI
 on:
   push:
-  branches:
-    - release
+    branches:
+      - release
 jobs:
   deploy:
     # A strategy is used to define various build environments this job will run.
@@ -11,7 +11,7 @@ jobs:
     # run on ubuntu, mac & windows.
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     # Runs when the commit message contains "[Released]"
     # if: "contains(github.event.head_commit.message, '[Released]')"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          mv bin/polkadot-archive/target/release/polkadot-archive ./release/polkadot-archive
+
+      - name: Get changelog
+        id: changelog_reader
+        uses: mindsers/changelog-reader-action@v2
+        with:
+          validation_depth: 2
+          path: ./CHANGELOG.md
```
emostov, line 46: How does it know which changelog entry? Do we need to put `version` here?

emostov (approved):
Look good from what I understand; just a question about how the changelog reader knows how to get the version. Is there a way to test without actually merging to the release branch?

.github/workflows/release.yml
insipx: > Look good from what I understand; just a question about how the changelog reader knows how to get the version. Is there a way to test without actually merging to the release branch?

Yeah, I tested this in my personal repo with the Substrate Archive Changelog: https://github.com/insipx/test-gh-action/releases/tag/v0.5.1

It will default to the latest version. The reason it knows is because this action knows the format of Keep A Changelog/it was made especially with that format in mind

RELEASE-CHECKLIST.md
```diff
@@ -5,24 +5,15 @@ Here's how to make a new release of `substrate-archive`.
...
-    1. Publish the release
-1. Signal to devops that there is a new release available
+2. Update `Cargo.toml` in `substrate-archive` and `substrate-archive-backend` with the new version number.
+3. Update all references to `substrate` crates and `polkadot` to their respective latest releases.
+4. Update the `CHANGELOG` as specified [here](https://keepachangelog.com/en/1.0.0/).
+5. Run all tests: `TEST_DATABASE_URL="postgres://localhost:5432/archive cargo test --all`.
+6. Push the PR against the `release` branch.
+7. Once reviewed, merge it to `release`. Upon merging, Github Actions will create a draft release and upload
+binaries.
+8. Tag the `release` branch` with `git tag vx.y.z` and push the tags with `git push --tags`.
```
maciejhirsz, line 15: Superfluous tick after `branch`.

maciejhirsz (approved):
One minor thing found.

RELEASE-CHECKLIST.md
TarikGul, line 15:
```suggestion
8. Tag the `release` branch with `git tag vx.y.z` and push the tags with `git push --tags`.
```

niklasad1 (approved):
looks cool.

PR #284, opened 2021-05-27
Update Dependencies

insipx:
- This fixes the compile for v0.9.3
- adds code_substitutes as option from the chain_spec


TarikGul (approved):
LGTM

substrate-archive-backend/src/frontend.rs
```diff
@@ -91,12 +91,13 @@ fn default_block_workers() -> usize {
...
 
-impl From<RuntimeConfig> for ClientConfig {
-	fn from(config: RuntimeConfig) -> ClientConfig {
+impl<B: BlockT> From<RuntimeConfig> for ClientConfig<B> {
+	fn from(config: RuntimeConfig) -> ClientConfig<B> {
 		ClientConfig {
 			offchain_worker_enabled: false,
 			offchain_indexing_api: false,
 			wasm_runtime_overrides: config.wasm_runtime_overrides,
+			wasm_runtime_substitutes: Default::default(),
```
dvdplm, line 100: This makes it compile I assume, but will it sync without the actual polkadot substitution when it gets to the infamous block?

dvdplm (commented):
I a bit concerned that using the default value for `wasm_runtime_substitutes` will work, did you try syncing past the troublesome block with this branch?

Another question I have is how `wasm_runtime_substitutes` interacts with `wasm_runtime_overrides`. Which one should trump the other when they overlap?

dvdplm:
Also: CI is angry.

substrate-archive-backend/src/frontend.rs
insipx: > Another question I have is how wasm_runtime_substitutes interacts with wasm_runtime_overrides. Which one should trump the other when they overlap?

Looking at the override code, code_substitute is used only if there isn't already an override passed in. So RuntimeOverrides trump code_substitute. 

To get over this hump when tracing is not enabled, however, will require some refactoring (working on that currently). Previously we didn't have to worry about the chain_spec at all when we could fully trust on-chain wasm. With `code_substitute` we will need to create ClientConfig out of the ChainSpec to sync past the problematic block. So yeah, `Default::default()` won't work in this case.

substrate-archive-backend/src/frontend.rs
```diff
@@ -73,6 +79,19 @@ pub struct RuntimeConfig {
 	pub wasm_pages: Option<u64>,
 	/// Path to WASM blobs to override the on-chain WASM with (required for state change tracing).
 	pub wasm_runtime_overrides: Option<PathBuf>,
+	/// code substitutes that should be used for the on chain wasm.
+	///
+	/// NOTE: Not to be confused with 'wasm_runtime_overrides'. code_substitutes
+	/// are included in the chain_spec and primarily for fixing problematic on-chain wasm.
```
dvdplm, line 85:
```suggestion
	/// NOTE: Not to be confused with 'wasm_runtime_overrides'. code_substitutes
	/// are included in the chain_spec and primarily for fixing problematic on-chain wasm. 
	/// If both are in use, the `was_runtime_overrides` takes precedence.
```

substrate-archive/src/archive.rs
```diff
@@ -53,16 +53,18 @@ pub struct ChainConfig {
 	pub(crate) rocksdb_secondary_path: Option<PathBuf>,
 	/// Chain spec.
 	#[serde(skip)]
-	pub(crate) spec: Option<ChainSpecConfig>,
+	pub(crate) spec: Option<Box<dyn ChainSpec>>,
```
dvdplm, line 56: Is the `spec` still optional after this?

substrate-archive/src/archive.rs
```diff
@@ -181,8 +183,7 @@ impl<B, R, D, DB> ArchiveBuilder<B, R, D, DB> {
 	}
 
 	/// Set the path to the secondary RocksDB database directory.
-	/// E.g. if you specify the path `./substrate-archive/rocksdb_secondary` and chain spec,
-	/// the actual path will be `./substrate-archive/rocksdb_secondary/<chain-spec-name>/<chain-spec-id>`.
+	/// E.g. if you specify the path `./substrate-archive/rocksdb_secondary` and chain spec, the actual path will be `./substrate-archive/rocksdb_secondary/<chain-spec-name>/<chain-spec-id>`.
```
dvdplm, line 186: long line

substrate-archive/src/archive.rs
insipx: I think it should remain optional in the archive lib, and then each individual library can enforce it if it needs to. Polkadot Archive must enforce that chain_spec exists because it uses `code_substitute` directly for instance, but another chain that does not rely on it won't have to.

Now that I consider it, it might make sense to just require chain spec regardless though

insipx:
@dvdplm @TarikGul Ran tests against live polkadot-archive node running on GCP. It was fairly obvious when I got the right wasm-blob because tracing wasm would stop producing anymore `State Root Mismatch`errors on every block, so I am confident this is ready for merge now that CI is passing :+1:

dvdplm (approved):

PR #288, opened 2021-06-02
Make release checklist more precise and add workflow for unit testing

insipx:

PR #289, opened 2021-06-02
chore(release): 0.5.2

TarikGul:


bin/node-template-archive/Cargo.toml
```diff
@@ -13,7 +13,7 @@ serde = "1.0"
 structopt = { version = "0.3", features = ["suggestions", "color"] }
 toml = "0.5"
 
-node-template-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
```
insipx, line 16: We need substrate-archive master to line up with substrate master

insipx:
Actually not sure why this diff is so large, I think `release`branch got screwed up. We shouldn't be so far ahead of master since we only cut one release out of this branch so far.

Drastic measure would be: 
- force-pushing (in order to rebase with `master`)
- deleting `release`and re-creating it 

we'd potentially have to fix 0.5.1 too. rel: https://github.com/paritytech/substrate-archive/pull/288

PR #292, opened 2021-06-09
Support Storage Chains

insipx:


insipx:
Not super happy with this kind of support because it feels like I copy-pasted the substrate code and modified it a bit. Would be nice if maybe things were more open in `client/db/src/lib`but it's understandable why they are not; but we need this kind of setup since we run an alternative `secondary-db` rather than a primary rocks instance (we have our own implementation for the database trait).

jhonalino:
this is the best thing ever. I've been pulling my hair out trying to query historical data via rpc and its super slow. 

I can't believe there is this substrate-archive project. this is the best thing ever

jhonalino:
thank you for making this project @insipx this exactly what I need!!

substrate-archive-backend/src/read_only_backend/blockchain_backend.rs
```diff
@@ -33,22 +34,59 @@ use sp_runtime::{
...
+			Some(body) => body,
+			None => return Ok(None),
+		};
+		match self.storage_mode {
+			TransactionStorageMode::BlockBody => match Decode::decode(&mut &body[..]) {
 				Ok(body) => Ok(Some(body)),
-				Err(_) => Err(BlockchainError::Backend("Could not decode extrinsics".into())),
+				Err(err) => return Err(BlockchainError::Backend(format!("Error decoding body: {}", err))),
+			},
+			TransactionStorageMode::StorageChain => match Vec::<ExtrinsicHeader>::decode(&mut &body[..]) {
```
dvdplm, line 61: I guess you could have a helper fn here, `fn decode_storage_chain_item<Block: BlockT, D: ReadOnlyDb>(body: &Vec<u8>, db: &D) -> ChainResult<Option<Vec<<Block as BlockT>::Extrinsic>>>` and sweep this under the rug with a `TransactionStorageMode::StorageChain => decode_storage_chain_item::<Block, D>(&body, &*self.db),`

Not convinced it's any better though. :/ Maybe it's better to keep it as closely aligned to substrate as possible if we need to copy&paste this.

Better still would be to use `sc_client_api::blockchain::Backend` directly, what was the problem there?

substrate-archive-backend/src/read_only_backend/blockchain_backend.rs
```diff
@@ -33,22 +34,59 @@ use sp_runtime::{
...
+				Ok(index) => {
+					let extrinsics: ChainResult<Vec<Block::Extrinsic>> = index
+						.into_iter()
+						.map(|ExtrinsicHeader { hash, data }| {
+							let decode_result = if hash != Default::default() {
+								match self.db.get(columns::TRANSACTION, hash.as_ref()) {
+									Some(t) => {
+										let input = [&data[..], &t[..]].concat();
```
dvdplm, line 69: Substrate has that funny `JoinInput` thing, I guess you went with this to cut down on dependencies? (Well, it's also not `pub` I guess).

I didn't read that code carefully, but you're confident concatenating is all we need here?

substrate-archive-backend/src/read_only_backend/blockchain_backend.rs
insipx: I think you're talking about `sc_client_db::Backend`?, in which case we could technically use that, but it wouldn't be very useful for archive so I made a design decision early on to form a custom Backend for Archive.

By 'not very useful' I mean that Archive uses some custom iteration logic to speed up gathering of it's blocks. For instance from our backend implementation we access the `iter` method on the database [here](https://github.com/paritytech/substrate-archive/blob/0df95bd6c5a661ab991d2e4900baff2276892877/substrate-archive-backend/src/read_only_backend.rs#L128). 

We could use Substrate's backend and compose it with our own backend, then re-implement the `Backend` traits on it to carry over it's functionality, but when I was first coding this stuff up that seemed like it was too complicated, and we would only use substrate Backend for backwards compatibility. There's also a bunch of references/stuff on substrate's backend that we wouldn't use pretty much at all in archive

substrate-archive-backend/src/read_only_backend/blockchain_backend.rs
insipx: Can confirm, added a test just to be sure it does the same thing

dvdplm (approved):

TarikGul (approved):
Lgtm, Question: Whats the main advantage of `TransactionStorageMode::StorageChain` vs `...::BlockBody` as a body scheme? I understand the core logic here, but I couldn't find any docs on the benefits or higher level reason for the `StorageChain`.

insipx:
> Lgtm, Question: Whats the main advantage of `TransactionStorageMode::StorageChain` vs `...::BlockBody` as a body scheme? I understand the core logic here, but I couldn't find any docs on the benefits or higher level reason for the `StorageChain`.

Best write-ups i've found are [here](https://github.com/paritytech/substrate/pull/7868) and the other PR's in the tracking issue [here](https://github.com/paritytech/substrate/issues/7962)


AFAIU it the storage chain feature comes down to chains that want to use a different storage scheme which allows them to point to specific pieces of stored extrinsics on-chain via something like IPFS. Hence why the extrinsics are stored in their own column in RocksDB

PR #294, opened 2021-06-11
suppress regalloc log messages

insipx:
Quick PR  to suppress these messages to `warn`. If not suppressed they seriously pollute the `Info` space and make it impossible to read anything that's going on

### DIFF
--- a/substrate-archive/src/logger.rs
+++ b/substrate-archive/src/logger.rs
@@ -75,6 +75,7 @@ pub fn init(config: LoggerConfig) -> io::Result<()> {
 		.level_for("cranelift_codegen", log::LevelFilter::Warn)
 		.level_for("header", log::LevelFilter::Warn)
 		.level_for("frame_executive", log::LevelFilter::Error)
+		.level_for("regalloc", log::LevelFilter::Warn)
 		.format(move |out, message, record| {
 			out.finish(format_args!(
 				"{} {} {}",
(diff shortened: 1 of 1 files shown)


niklasad1 (approved):

substrate-archive/src/logger.rs
```diff
@@ -98,6 +99,7 @@ pub fn init(config: LoggerConfig) -> io::Result<()> {
 			.level_for("staking", log::LevelFilter::Warn)
 			.level_for("cranelift_codegen", log::LevelFilter::Warn)
 			.level_for("wasm-heap", log::LevelFilter::Error)
+			.level_for("regalloc", log::LevelFilter::Warn)
```
dvdplm, line 102: dq: what is the `regalloc` module?

dvdplm (approved):

substrate-archive/src/logger.rs
niklasad1: it's some `CPU instruction register` allocation crate that is used by `wasm/wasmtime` but not sure how it's used in archive, maybe @insipx  can explain :)

insipx: It's not used in archive at all actually. During one of the 'update dependency' PR's, something must have changed within `regalloc` itself, or maybe the wasm executor in substrate, but it was printing out basically every instruction going through the executor which made it impossible to read any other logs

PR #301, opened 2021-06-22
Fix Shutdown for Archive

insipx:
To be merged after #254 

- Greatly simplify 'System'
  - System is made into a Factory that spawns 'SystemInstances'
- Shutdown takes max ~1 second now. 
- Switches out `smol` for `async-std`. Both `smol` and `async-std` use the same libraries. However the usage of smol in archive was becoming extensive,  and it would've required creating a more extensive executor (using smols `Executor`) and exercising detailed control over how the threads are spawned etc. I decided that we didn't really need that much control and it would be easier to just use `async-std`.


substrate-archive/src/actors/workers/database.rs
```diff
@@ -102,7 +102,7 @@ impl DatabaseActor {
 		let len = block_nums.len();
 		let now = std::time::Instant::now();
 		while queries::has_blocks(block_nums.as_slice(), &mut conn).await?.len() != len {
-			smol::Timer::after(std::time::Duration::from_millis(50)).await;
+			Delay::new(std::time::Duration::from_millis(50)).await;
```
niklasad1, line 105: why don't use https://docs.rs/async-std/1.9.0/async_std/future/fn.timeout.html instead?

niklasad1 (approved):
LGTM, nice

TarikGul (approved):
lgtm

PR #309, opened 2021-06-27
[WIP] Add block justification support

### LINKED ISSUE #147: Store Justifications along with the rest of the block
For one reason or another Justifications aren't stored in the Postgres Database. 

Justification is present on the `SignedBlock` field of `Block` though, so it would be nice to store it if it exists. This requires a new SQL migration that creates a `Justification` column as a `bytea` field. It is possible for justification to be null for a block, so the columns should reflect this.

jakehemmerle:
waiting on a blocking sqlx bug mentioned in #147 

should close #147


substrate-archive/src/database/models.rs
```diff
@@ -57,6 +58,8 @@ impl BlockModel {
 		let header = <B::Header as HeaderT>::new(block_num, extrinsics_root, state_root, parent_hash, digest);
 
 		let spec = self.spec as u32;
+		// do I need to include justifications here?
```
insipx, line 61: I'd say they're not necessary here, because we don't need the justification to build the block / execute the block later on.

substrate-archive/src/migrations/20210611075511_add_column_justification.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+alter table blocks
```
insipx, line 2:
```suggestion
ALTER TABLE blocks
```

substrate-archive/src/migrations/20210611075511_add_column_justification.sql
```diff
@@ -0,0 +1,3 @@
...
+	add justification bytea;
```
insipx, line 3:
```suggestion
	ADD justification bytea;
```

insipx (commented):
looks good so far! hopefully the SQLx bug gets fixed :)

PR #311, opened 2021-06-30
set default log file name = current time in UTC

### LINKED ISSUE #304: Start a new log file if logfile already detected
On startup we should create a unique logfile with the current UTC time.

jakehemmerle:
example filename is `2021-06-30T17:08:24.692011+00:00.log`. manually tested. should close #304

Any way to make this a 1-liner? I tried `chrono::Utc::now().to_rfc3339().push_str(".log")` but I forgot `push_str` returns `()` instead of our cool new timestamped log file name.


dvdplm (approved):
Using two lines is idiomatic rust (but I see where you're coming from, I've had the same thought).

insipx (approved):

PR #315, opened 2021-07-13
fix listener, the bug causing unit tests to fail

insipx:
Fixes a bug with listener where if the listener is spawned and then immediately issued notifications, the listener might miss the first few messages.

In addition I modified the listener `kill()` function to ensure the task finishes on kill (waits for task to be joined after sending the message to close).

### DIFF
--- a/substrate-archive/src/actors.rs
+++ b/substrate-archive/src/actors.rs
@@ -261,7 +261,7 @@ where
 	B::Header: serde::de::DeserializeOwned,
 {
 	fn new(config: SystemConfig<B, D>, client: Arc<C>) -> Result<Self> {
-		let listener = Self::init_listeners(&config)?;
+		let listener = task::block_on(Self::init_listeners(&config))?;
 		Ok(Self { config, client, listener, _marker: PhantomData })
 	}
 
@@ -298,7 +298,7 @@ where
 		Ok(())
 	}
 
-	fn init_listeners(config: &SystemConfig<B, D>) -> Result<Listener> {
+	async fn init_listeners(config: &SystemConfig<B, D>) -> Result<Listener> {
 		Listener::builder(config.pg_url(), move |notif, conn| {
 			async move {
 				let sql_block = queries::get_full_block_by_number(conn, notif.block_num).await?;
@@ -310,6 +310,7 @@ where
 		})
 		.listen_on(Channel::Blocks)
 		.spawn()
+		.await
 	}
 
 	/// Checks if any blocks that should be executed are missing
--- a/substrate-archive/src/actors.rs
+++ b/substrate-archive/src/actors.rs
@@ -335,7 +335,9 @@ where
 
 impl<B, R, D, C> Drop for SystemInstance<B, R, D, C> {
 	fn drop(&mut self) {
-		task::block_on(async { self.listener.kill().await })
+		if let Err(e) = task::block_on(async { self.listener.kill().await }) {
+			log::error!("{}", e)
+		}
 	}
 }
 
--- a/substrate-archive/src/actors.rs
+++ b/substrate-archive/src/actors.rs
@@ -335,7 +335,7 @@ where
 
 impl<B, R, D, C> Drop for SystemInstance<B, R, D, C> {
 	fn drop(&mut self) {
-		if let Err(e) = task::block_on(async { self.listener.kill().await }) {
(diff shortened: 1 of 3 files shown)


insipx:
Tests fail b/c:
```sh
 Error: failed to download from `https://crates.io/api/v1/crates/async-mutex/1.4.0/download`
```
going to re-run the tests tomorrow morning EST assuming something is funky between github/crates.io

substrate-archive/src/database/listener.rs
```diff
@@ -122,13 +124,19 @@ where
...
-		let fut = async move {
+		// NOTE: this part is not included in the main future in order to prevent missing messages.
+		// Otherwise, it would be possible to spawn, immediately send a notification, which would be missed if we are not connected/listening yet.
+		let listener = task::block_on(async {
 			let mut listener = PgListener::connect(&pg_url).await?;
 			let channels = self.channels.iter().map(String::from).collect::<Vec<String>>();
 			listener.listen_all(channels.iter().map(|s| s.as_ref())).await?;
-			let mut conn = PgConnection::connect(&pg_url).await.unwrap();
+			Ok::<PgListener, ArchiveError>(listener)
+		})?;
```
jsdw, line 134: This looks good to me.

I'm just wondering; instead of blocking the current thread on `spawn`, would it make sense at some point to make `spawn` an `async fn` and await it in order to guarantee that it's ready before anything else happens?

substrate-archive/src/database/listener.rs
```diff
@@ -158,22 +166,24 @@ where
...
-						self.handle_listen_event(msg.unwrap(), &mut conn).await;
-					}
+			let gather_unfinished = || async {
+				let notifs = listener.collect::<Vec<_>>().await;
+				for msg in notifs {
+					self.handle_listen_event(msg?, &mut conn).await;
 				}
+				Ok::<(), ArchiveError>(())
+			};
+			if timeout(Duration::from_secs(1), gather_unfinished()).await.is_err() {
```
jsdw, line 179: Much nicer!

jsdw (approved):
LGTM! Just a question about the task blocking :)

substrate-archive/src/actors.rs
```diff
@@ -335,7 +335,9 @@ where
 
 impl<B, R, D, C> Drop for SystemInstance<B, R, D, C> {
 	fn drop(&mut self) {
-		task::block_on(async { self.listener.kill().await })
+		if let Err(e) = task::block_on(async { self.listener.kill().await }) {
```
dvdplm, line 338: DQ: why couldn't this be `task::block_on(self.listener.kill())`?

substrate-archive/src/database/listener.rs
```diff
@@ -158,22 +166,24 @@ where
...
-				_ = FutureExt::fuse(timeout) => {},
-				notifs = listener.collect::<Vec<_>>().fuse() => {
-					for msg in notifs {
-						self.handle_listen_event(msg.unwrap(), &mut conn).await;
-					}
+			let gather_unfinished = || async {
+				let notifs = listener.collect::<Vec<_>>().await;
+				for msg in notifs {
+					self.handle_listen_event(msg?, &mut conn).await;
 				}
```
dvdplm, line 176: Could you use `listener.for_each()` here you think?

dvdplm (approved):
LGTM, left a few questions.

substrate-archive/src/actors.rs
insipx: Oh it totally can, good catch

substrate-archive/src/database/listener.rs
insipx: Tried this but because we need a `&mut reference` for the Postgres connection so we can't use that inside of a closure here + it's easier to handle the errors this way (for_each would force us to return a `()`)

substrate-archive/src/database/listener.rs
insipx: Yeah that would make sense, I think I had it that way in the past at some point in the past.

I transitioned to making the spawn blocking because it simplifies the startup, and by not making it async we can maintain a handle to the listener in the main `SystemInstance` object, which then allows us to clean it up on `Drop`. However it is very context specific and it probably would make more sense from an API standpoint to await this part and then use block_on if necessary outside of instantiation.

PR #390, opened 2021-11-14
Update rust 2021, MSRV 1.56.1, `authors = ["Parity Technologies <admin@parity.io>"]`

NukeManDan:
missed one ;)

### DIFF
--- a/bin/node-template-archive/Cargo.toml
+++ b/bin/node-template-archive/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "node-template-archive"
 version = "0.3.0"
-authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
+authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
 rust-version = "1.56.1"
 description = "Indexes Substrate Data"
--- a/bin/node-template-archive/Cargo.toml
+++ b/bin/node-template-archive/Cargo.toml
@@ -3,6 +3,7 @@ name = "node-template-archive"
 version = "0.3.0"
 authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
 edition = "2021"
+rust-version = "1.56.1"
 description = "Indexes Substrate Data"
 
 [dependencies]
--- a/substrate-archive/src/actors.rs
+++ b/substrate-archive/src/actors.rs
@@ -199,6 +199,7 @@ where
 		self.blocks.send(ReIndex).await?;
 		let actors = self.clone();
 		task::spawn(async move {
+			let _ = &actors;
 			loop {
 				let fut = (
 					Box::pin(actors.blocks.send(Crawl)),
--- a/substrate-archive/src/actors.rs
+++ b/substrate-archive/src/actors.rs
@@ -199,7 +199,6 @@ where
 		self.blocks.send(ReIndex).await?;
 		let actors = self.clone();
 		task::spawn(async move {
-			let _ = &actors;
 			loop {
 				let fut = (
 					Box::pin(actors.blocks.send(Crawl)),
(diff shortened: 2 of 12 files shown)


substrate-archive/src/actors.rs
```diff
@@ -199,6 +199,7 @@ where
 		self.blocks.send(ReIndex).await?;
 		let actors = self.clone();
 		task::spawn(async move {
+			let _ = &actors;
```
niklasad1, line 202: remove?!

```suggestion
```

NukeManDan: this is from cargo fix --edition. I would say refactoring or leaving is best, as I think this pattern saves us from a breaking change planned in behavior in this edition

NukeManDan:
MSRV is The "minimum supported Rust version" right?

"increments are a breaking Semver change" policy may applies to this repository 🤔 should it?

> The relevant check should update to the minimal dependency versions first, in order to make sure newer compatible versions are ignored in this regard.

AFAIK minimal stable toolchain version with rust edition 2021 is rustc 1.56.0
https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html

This could be propagated in all change logs and maybe include a version bump. But what type? patch, minor or major?
Or no bump?

substrate-archive/src/actors.rs
niklasad1: it's not a breaking change rather that the reference is unused thus it's a bug in `cargo fix` FWIW :)

//cc @insipx  ^^

insipx: Yeah i think removing this is the play. Compiles just fine w/o it

insipx (approved):

insipx:
> MSRV is The "minimum supported Rust version" right?
> 
> "increments are a breaking Semver change" policy may applies to this repository thinking should it?
> 
> > The relevant check should update to the minimal dependency versions first, in order to make sure newer compatible versions are ignored in this regard.
> 
> AFAIK minimal stable toolchain version with rust edition 2021 is rustc 1.56.0 https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html
> 
> This could be propagated in all change logs and maybe include a version bump. But what type? patch, minor or major? Or no bump?

We should just do a minor version bump for this, maybe include it in the 0.7 release with the upcoming extrinsic decoding functionality. If you'd like you can add a line to the current CHANGELOG under 'unreleased', otherwise i'll go through and put in all the changes on next release :)

rust 1.56 seems fine to me

NukeManDan:
`rust-version = "1.56.1"` introduced

For reference, I tried to get this to `cargo build` with 1.56.0 and it failed with
```
error: package `...` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
```
With `--ignore-rust-version` it _does_ build with 1.56.0, as expected.

bin/node-template-archive/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "node-template-archive"
 version = "0.3.0"
 authors = ["Andrew Plaza <aplaza@liquidthink.net>"]
```
niklasad1, line 4: unrelated to this PR: @insipx we should probably as `parity` as authors too to all crates?!

niklasad1 (approved):

bin/node-template-archive/Cargo.toml
NukeManDan: happy to use whatever, you feel free to commit in this PR's branch too whatever you want. perhaps: 
 https://github.com/paritytech/substrate/blob/master/bin/node/runtime/Cargo.toml#L4
 ```toml
 authors = ["Parity Technologies <admin@parity.io>"]
 ```

NukeManDan: updated the last commit to use this throughout, please `git reset HEAD~ --hard` if you think it's not appropriate in this PR (new title)

NukeManDan:
Think this is good to merge? @insipx if you think so, squash it in at your convenience :)

insipx (approved):

PR #91, opened 2020-09-02
[WIP] Preliminary integration with desub

insipx:
Storage decoding has succesfully been implemented in [desub](https://github.com/paritytech/desub). This PR adds desub support to substrate-archive in the form of decoded extrinsics, and a new schema that provides a table-per-pallet for storage.

This PR will redirect the destination of outputs from the `Metadata` Actor to a `Decoder` actor for blocks, and from the `StorageAggregator` actor to the `Decoder` actor for storage.

this is still in testing/experimental and I sort of expect many things to go wrong

PR #394, opened 2021-11-16
Extrinsic Decoding

insipx:
Mostly done, running against chains to check performance and observing how it is to query against older and newer JSON for specs that use older/newer metadata

this also updates archive to use latest substrate

- [x] handle upgrade blocks
- [x] desub track master? (avoid dup dependencies)

supersedes #91


substrate-archive/src/actors/workers/extrinsics_decoder.rs
```diff
@@ -0,0 +1,139 @@
...
+
+		let versions: Vec<u32> =
+			blocks.iter().filter(|b| !self.decoder.has_version(&b.3)).map(|(_, _, _, v)| *v).unique().collect();
+		// above and below line are separate to let immutable ref to `self.decoder` to go out of scope.
+		for version in versions.iter() {
+			let metadata = queries::metadata(&mut conn, *version as i32).await?;
+			log::debug!("Registering version {}", version);
+			// TODO: FIX EXPECT
+			Arc::get_mut(&mut self.decoder)
+				.expect("Actors guarantee one reference; qed")
```
insipx, line 72: This doesn't necessarily hold true, i should make expect into error or remove arc and use a different solution

substrate-archive/src/database/queries.rs
```diff
@@ -194,6 +221,57 @@ pub(crate) fn blocks_paginated<'a>(
...
+			FROM blocks
+			WHERE spec != 0
+			ORDER BY spec, block_num ASC
+		"#,
+	)
+	.fetch_all(conn)
+	.await?
+	.into_iter()
+	.tuple_windows()
+	.filter(|(_curr, next)| next.spec >= from)
```
insipx, line 268: i tried doing this in postgres and 'returning one row previous from the `from` spec, but failed to figure that out. This is simpler/easier way to do it

substrate-archive/src/actors/workers/extrinsics_decoder.rs
```diff
@@ -0,0 +1,144 @@
...
+
+	fn decode(
+		decoder: &Decoder,
+		blocks: Vec<(u32, Vec<u8>, Vec<u8>, u32)>,
+		upgrades: &HashMap<u32, u32>,
+	) -> Result<Vec<ExtrinsicsModel>> {
+		let mut extrinsics = Vec::new();
+		for (number, hash, ext, spec) in blocks.into_iter() {
+			if let Some(version) = upgrades.get(&number) {
+				let previous = itertools::sorted(upgrades.values())
```
dvdplm, line 99: Why is clippy so upset about `itertools` not being used?

substrate-archive/src/database/models.rs
```diff
@@ -157,11 +178,13 @@ pub struct PersistentConfig {
 	pub minor: i32,
 	/// Patch version of this library.
 	pub patch: i32,
+	/// The chain data this db is populated with
+	pub chain: String,
```
dvdplm, line 182: We should probably use genesis hashes for this rather than the chain name. Or a `(GenesisHash, String)` tuple to have the string for logs.

substrate-archive/src/database/queries.rs
```diff
@@ -194,6 +223,57 @@ pub(crate) fn blocks_paginated<'a>(
 	})
 }
 
+/// Get up to `max_block_load` extrinsics which are not present in the `extrinsics` table.
+pub(crate) async fn missing_extrinsic_blocks(
```
dvdplm, line 227: Maybe `blocks_missing_extrinsics` is a better name for this?

substrate-archive/src/database/queries.rs
dvdplm: Possibly actually faster too.

substrate-archive/src/types.rs
```diff
@@ -144,6 +146,25 @@ impl<Hash: Send + Sync + 'static> Message for BatchStorage<Hash> {
...
+}
+
+impl BatchExtrinsics {
+	pub fn new(extrinsics: Vec<ExtrinsicsModel>) -> Self {
+		Self { inner: extrinsics }
+	}
+
+	pub fn inner(self) -> Vec<ExtrinsicsModel> {
+		self.inner
+	}
```
dvdplm, line 161: Maybe a `len` method is useful to add here?

dvdplm (approved):
lgtm, left a few nits. CI is pretty upset though?

dvdplm:
Building locally I get:

```
rror: error returned from database: relation "extrinsics" does not exist
   --> substrate-archive/src/database/queries.rs:231:15
    |
231 |       let blocks = sqlx::query_as!(
    |  __________________^
232 | |         BlockExtrinsics,
233 | |         "
234 | |         SELECT block_num, hash, ext, spec FROM blocks
...   |
240 | |         i64::from(max_block_load)
241 | |     )
    | |_____^
    |
    = note: this error originates in the macro `$crate::sqlx_macros::expand_query` (in Nightly builds, run with -Z macro-backtrace for more info)
```

What am I doing wrong?

jsdw:
@dvdplm Hmm, when I try to do a local run (eg with `cargo check` it seems to pick up the offline `substrate-archive/sqlx-data.json` file. Perhaps you have a `DATABASE_URL` env var (which overwrites the default offline mode in checks)? You could try adding the `SQLX_OFFLINE=true` env var to counteract that.

It does look like the offline definitions aren't uptodate yet though; I get:

```
error: failed to find data for query 9e6316290793ef9ca02c1a917d4bd3412497336b1928222381591f2a7a00e3cb at line 206 column 1
   --> substrate-archive/src/database/queries.rs:130:2
    |
130 |     sqlx::query_as!(Meta, "SELECT meta FROM metadata WHERE version = $1", spec)
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the macro `$crate::sqlx_macros::expand_query` (in Nightly builds, run with -Z macro-backtrace for more info)

error: failed to find data for query eafd9a045fa56835703150bd9ad24801207e0326287cee132f45a5273d3f3200 at line 206 column 1
   --> substrate-archive/src/database/queries.rs:231:15
    |
231 |       let blocks = sqlx::query_as!(
    |  __________________^
232 | |         BlockExtrinsics,
233 | |         "
234 | |         SELECT block_num, hash, ext, spec FROM blocks
...   |
240 | |         i64::from(max_block_load)
241 | |     )
    | |_____^
    |
    = note: this error originates in the macro `$crate::sqlx_macros::expand_query` (in Nightly builds, run with -Z macro-backtrace for more info)

error: failed to find data for query 05b056299b2e00d9d96eb6357329e8fdb77aede4485e045bb71d75c1e6feba1e at line 206 column 1
   --> substrate-archive/src/database/queries.rs:257:15
    |
257 |       let blocks = sqlx::query_as!(
    |  __________________^
258 | |         BlockNumSpec,
259 | |         r#"
260 | |             SELECT DISTINCT ON (spec) spec, block_num
...   |
264 | |         "#,
265 | |     )
    | |_____^
    |
    = note: this error originates in the macro `$crate::sqlx_macros::expand_query` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I assume those queries are new/updated :)

substrate-archive/src/actors.rs
```diff
@@ -204,8 +211,9 @@ where
 					Box::pin(actors.blocks.send(Crawl)),
 					Box::pin(actors.storage.send(SendStorage)),
 					Box::pin(actors.storage.send(SendTraces)),
+					Box::pin(actors.extrinsics.send(Index)),
 				);
-				if let (Err(_), Err(_), Err(_)) = future::join3(fut.0, fut.1, fut.2).await {
+				if let (Err(_), Err(_), Err(_), Err(_)) = future::join4(fut.0, fut.1, fut.2, fut.3).await {
```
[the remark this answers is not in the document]
jsdw: Good catch! Offhand I'd assume that we'd want any error to break, but perhaps the intention is to keep looping around until everything succeeds, unless everything fails?

I'm not sure!

substrate-archive/src/database/models.rs
```diff
@@ -219,6 +258,17 @@ impl PersistentConfig {
 		let version = semver::Version::parse(version)?;
 		Ok((version.major.try_into()?, version.minor.try_into()?, version.patch.try_into()?))
 	}
+
+	pub(crate) fn chain(&self) -> Chain {
+		match self.chain.to_ascii_lowercase().as_str() {
+			"kusama" => Chain::Kusama,
+			"polkadot" => Chain::Polkadot,
+			"westend" => Chain::Westend,
+			"centrifuge" => Chain::Centrifuge,
```
jsdw, line 267: I haven't heard of centrifuge before!

substrate-archive/src/database/queries.rs
jsdw: I had to try this, and the best I have is:

```
WITH 
  exp AS (
    SELECT DISTINCT ON (spec) spec, block_num FROM blocks WHERE spec != 0 ORDER BY spec
  ),
  lagged AS (
    SELECT spec, block_num, LAG(block_num, 1, 0) OVER (ORDER BY spec) as prev_block_num FROM exp
  )
SELECT * FROM lagged WHERE block_num = 314201;
```

I couldn't really understand why, but if I put the `LAG` straight into the `lagged` expr it was all fine until I applied a WHERE clause, and then nothing.... but anyway.. yes, your approach is simpler!

insipx:
yeah I haven't generated sqlx_offline yet which is why the queries aren't compiling

Also working on this cleanup, thanks for the reviews so far!

substrate-archive/src/database/models.rs
dvdplm: aka Altair on Kusama. :)

substrate-archive/src/database/queries.rs
insipx: Yeah I got about as far, with the same issue -- as soon as `WHERE`is added it filters out the previous spec

Thankfully not too big a deal though

substrate-archive/src/database/models.rs
insipx: Yup, Altair

they're defined here because they also exist directly in PolkadotJS defs too

substrate-archive/src/actors.rs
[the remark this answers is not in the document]
insipx: Yeah the only error that can occur is a `Disconnect`from `xtra`, which typically happens on shutdown and all them disconnect. Changed to `try_join`

substrate-archive/src/actors/workers/extrinsics_decoder.rs
insipx: :shrug: 
i changed it to not use itertools::sorted but instead .sorted() from the trait -- maybe that helped?

substrate-archive/src/database/models.rs
insipx: added genesis_hash column

substrate-archive/src/database/models.rs
jsdw: Aah ok, that's handy to know! Thanks!

insipx:
Troubleshooting a rather extreme memory issue with this PR, so it shouldn't be merged yet

substrate-archive/src/actors/workers/extrinsics_decoder.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Crawls missing extrinsics upon receiving an `Index` message.
+pub struct ExtrinsicsDecoder {
+	/// Pool of Postgres Connections.
+	pool: PgPool,
+	/// Address of the database actor.
+	addr: Address<DatabaseActor>,
+	/// Max amount of extrinsics to load at any one time.i.
+	max_block_load: u32,
+	/// Desub Legacy + current decoder.
+	decoder: Arc<Decoder>,
```
jsdw, line 46: At the moment, it's not clear to me why we need to surround this (and below) in Arc? As far as I can see (very much offhand and I'm almost certain I'm missing something at the mo) this can't be cloned, and is probably a single long running actor that doesn't need synchronisation things?

jsdw: Aah I see; it's for spawning a copy of this onto a task below; got it!

substrate-archive/src/migrations/20211108182913_extrinsics.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS extrinsics (
+	id SERIAL NOT NULL,
+	hash bytea NOT NULL PRIMARY KEY,
+	number int check (number >= 0 and number < 2147483647) NOT NULL UNIQUE,
```
jsdw, line 4: I guess the `< 2147483647` is not necessary since int is signed 4 bytes, but there's something nice about documenting the range explicitly anyway, since you need the `>=0` bit :)

jsdw: Second Q: Should `number` be a foreign key pointing to `blocks.block_num`?

substrate-archive/src/database/queries.rs
```diff
@@ -194,6 +231,82 @@ pub(crate) fn blocks_paginated<'a>(
...
+pub(crate) async fn blocks_missing_extrinsics(
+	conn: &mut PgConnection,
+	max_block_load: u32,
+) -> Result<Vec<(u32, Vec<u8>, Vec<u8>, u32)>> {
+	let blocks = sqlx::query_as!(
+		BlockExtrinsics,
+		"
+		SELECT block_num, hash, ext, spec FROM blocks
+		WHERE NOT EXISTS
+			(SELECT number FROM extrinsics WHERE extrinsics.number = blocks.block_num)
```
jsdw, line 245: I wonder how this compares in performance in contrast to something like:

```
SELECT block_num, hash, ext, spec FROM blocks
LEFT OUTER JOIN extrinsics ON extrinsics.number = blocks.block_num
WHERE extrinsics.number IS NULL
```

(I 've no real reason to think that they wouldn't resolve to an identical plan; I'm just more used to using joins for this sort of thing :))

substrate-archive/src/database/queries.rs
```diff
@@ -219,6 +332,7 @@ mod tests {
 
 	// Setup a data scheme such that:
 	// - blocks 0 - 200 will be missing from the `storage` table
+	// - blocks 0-400 will be missing from the `extrinsics` table
```
jsdw, line 335: Total nit: inconsistent spacing around dash :D

substrate-archive/src/lib.rs
```diff
@@ -15,7 +15,7 @@
 // along with substrate-archive.  If not, see <http://www.gnu.org/licenses/>.
 
 #![forbid(unsafe_code)]
-#![deny(dead_code)]
+// #![deny(dead_code)]
```
jsdw, line 18: Nit: remove permanently?

jsdw (approved):
I had a skim over; didn't try actually running it though! LGTM offhand :)

substrate-archive/src/migrations/20211108182913_extrinsics.sql
jsdw: That's fair! Def happy with it merging as-is for now :)

substrate-archive/src/database/queries.rs
insipx: hm, the query in archive is faster, but tbth i don't really know why, query planning in postgres is a bit of an enigma to me, i just wrote it this way because it's what i'm used to. fwiw here are the plans for both (starting with archive)

```sql
SELECT block_num, hash, ext, spec FROM blocks
WHERE NOT EXISTS
	(SELECT number FROM extrinsics WHERE extrinsics.number = blocks.block_num)
ORDER BY block_num ASC
```

```sql
"Gather Merge  (cost=1043.17..1091237.53 rows=433043 width=140) (actual time=1921.984..1984.607 rows=371882 loops=1)"
"  Workers Planned: 2"
"  Workers Launched: 2"
"  ->  Merge Anti Join  (cost=43.14..1040253.60 rows=180435 width=140) (actual time=1900.154..1932.888 rows=123961 loops=3)"
"        Merge Cond: (blocks.block_num = extrinsics.number)"
"        ->  Parallel Index Scan using blocks_block_num_index on blocks  (cost=0.43..614785.20 rows=3234289 width=140) (actual time=0.016..566.381 rows=2591749 loops=3)"
"        ->  Index Only Scan using extrinsics_number_key on extrinsics  (cost=0.43..368521.01 rows=7329251 width=4) (actual time=0.040..892.288 rows=7403366 loops=3)"
"              Heap Fetches: 4274634"
"Planning Time: 0.175 ms"
"JIT:"
"  Functions: 15"
"  Options: Inlining true, Optimization true, Expressions true, Deforming true"
"  Timing: Generation 0.990 ms, Inlining 46.963 ms, Optimization 69.031 ms, Emission 47.432 ms, Total 164.417 ms"
"Execution Time: 1993.836 ms"
``` 




```sql
EXPLAIN ANALYZE
SELECT block_num, blocks.hash, ext, spec FROM blocks
LEFT OUTER JOIN extrinsics ON extrinsics.number = blocks.block_num
WHERE extrinsics.number IS NULL
```

```sql
"Gather  (cost=376870.23..1076040.01 rows=433043 width=140) (actual time=1697.753..2332.745 rows=371882 loops=1)"
"  Workers Planned: 2"
"  Workers Launched: 2"
"  ->  Parallel Hash Anti Join  (cost=375870.23..1031735.71 rows=180435 width=140) (actual time=1676.099..2080.404 rows=123961 loops=3)"
"        Hash Cond: (blocks.block_num = extrinsics.number)"
"        ->  Parallel Seq Scan on blocks  (cost=0.00..489912.89 rows=3234289 width=140) (actual time=0.035..452.304 rows=2591749 loops=3)"
"        ->  Parallel Hash  (cost=325767.05..325767.05 rows=3053855 width=4) (actual time=698.715..698.715 rows=2467789 loops=3)"
"              Buckets: 131072  Batches: 128  Memory Usage: 3328kB"
"              ->  Parallel Index Only Scan using extrinsics_number_key on extrinsics  (cost=0.43..325767.05 rows=3053855 width=4) (actual time=0.020..404.442 rows=2467789 loops=3)"
"                    Heap Fetches: 1424878"
"Planning Time: 0.175 ms"
"JIT:"
"  Functions: 21"
"  Options: Inlining true, Optimization true, Expressions true, Deforming true"
"  Timing: Generation 1.344 ms, Inlining 46.064 ms, Optimization 39.578 ms, Emission 21.892 ms, Total 108.879 ms"
"Execution Time: 2341.940 ms"
```

the left join ends up doing a sequential scan on `blocks` which is probably why it's slower

substrate-archive/src/lib.rs
insipx: meant to uncomment :+1:

substrate-archive/src/database/queries.rs
dvdplm: I've seen subqueries be faster like this before, which always struck me as a  bit magical. Joins, being "harder" to write and think through feels like they should be faster. :)

jsdw: Ooh, very interesting! It's definitely a bit of a dark art to me, too. 

I'm surprised that such different strategies were picked; I wonder whether the `ORDER BY block_num ASC` had an impact?

I do like the subquery more, so it's good to see that it's actually more performant too in this case :)

PR #405, opened 2021-12-13
migrate before updating

insipx:


dvdplm (approved):
Clippy seems angry though.

PR #438, opened 2022-02-12
add nix support for dev

m0ssc0de:


bernardoaraujor:
I ran a successful build with the proposed `shell.nix` on my NixOS setup, which probably means the PR is ok.

I'm a bit confused about the dependencies though.
Given the [prerequisites](https://github.com/paritytech/substrate-archive#prerequisites), why is it enough that `shell.nix` has no mentions to PostgresSQL?

ggwpez:
> I'm a bit confused about the dependencies though. Given the [prerequisites](https://github.com/paritytech/substrate-archive#prerequisites), why is it enough that `shell.nix` has no mentions to PostgresSQL?

Maybe for running the tests?  
I think if we want to to fully support this we should add a CI build+test step.

bernardoaraujor:
> > I'm a bit confused about the dependencies though. Given the [prerequisites](https://github.com/paritytech/substrate-archive#prerequisites), why is it enough that `shell.nix` has no mentions to PostgresSQL?
> 
> Maybe for running the tests?  
> I think if we want to to fully support this we should add a CI build+test step.

CI build+test is indeed important for long term sustainability.

I submitted a similar `shell.nix` PR to `substrate-contracts-node` recently but the ink! team rejected it based on the maintenance burden it would add to them.

Perhaps the code owner of this project should be on the loop.

dvdplm:
> Perhaps the code owner of this project should be on the loop.

We're here reading. Not `nix` users ourselves (yet) so you are indeed correct that this would have to be community maintained and it is a concern. I'm not opposed to adding it if we see active users step up.

insipx:
This looks good, I can post my old nix config too if that's of interest. I think [Oxalica](https://github.com/oxalica/rust-overlay) is definitely better for managing rust dependencies IMO.

Issue #431, opened 2022-01-25
Make AMQP Url configurable via Environment variables


insipx:
the rabbitMQ url should have the possibility of being configured via the environment similar to how `DATABASE_URL` and `CHAIN_DATA_DB` are. The environment variable could be `AMQP_URL`

joaoepj:
Hi @insipx, I have introduced the AMQP_URL variable into my fork.
1. I've created a function env_var_task_url() that uses AMQP_URL value, or fallback to default_task_url(). Is this enough, or should I create some ControlConfig implementation?
2. Currently there's no AMQP_URL validation, which level of validation is expected here?

insipx:
1.) this sounds good; there is already a control config implementation for the task url: https://github.com/paritytech/substrate-archive/blob/eeb593ed13867642c9b80a39556a75a36c4e47bc/substrate-archive/src/actors.rs#L100

but the environment variable should take precedence over that option

2.) if the URL cannot connect the amqp library (lapin) would throw an error; otherwise i'm unsure of what kind of validation would be appropriate here. I would like to say that just accepting any URL would be fine

PR #452, opened 2022-03-24
Add AMQP_URL environment variable

joaoepj:
This PR enables setting the value of ControlConfig::task_url()
 via the AMQP_URL encironment variable. It relates to [issue 431](https://github.com/paritytech/substrate-archive/issues/431)

### DIFF
--- a/bin/polkadot-archive/src/main.rs
+++ b/bin/polkadot-archive/src/main.rs
@@ -31,6 +31,8 @@ pub fn main() -> Result<()> {
 	let cli = cli_opts::CliOpts::init();
 	let config = cli.parse()?;
 
+
+	
 	let mut archive = run_archive::<SecondaryRocksDb>(&cli.chain_spec, config)?;
 	archive.drive()?;
 	let running = Arc::new(AtomicBool::new(true));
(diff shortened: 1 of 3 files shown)


insipx:
this is OK, but the config file will always override the environment variable. I think it would make more sense to put the check for the environment variable along with the env variables `CHAIN_DATA_DB` and `DATABASE_URL` here: https://github.com/paritytech/substrate-archive/blob/6e5a6e2d22d7dd28bfb4f2eb50e5f2de6045e433/substrate-archive/src/archive.rs#L350

and then replace config.control.task_url from there if it exists. this way env variable always takes precedence over the config file

joaoepj:
@insipx Ops... slowly getting used to rust basics and beginning to understand something. As you suggested I checked the AMQL_URL along with other environment variables in ArchiveBuilder::build() and it also ended up with file config prevailing over env config. So I added another check in ArchiveBuilder::with_config() achieving the desired result. Seems like ArchiveBuilder::build() scenario  is used to do some database recovery. I don't figured it out exactly but leave a duplicated check there anyway. Let me know what else is needed polish it up.
 
> this is OK, but the config file will always override the environment variable. I think it would make more sense to put the check for the environment variable along with where the check occurs for env variables `CHAIN_DATA_DB` and `DATABASE_URL` here:
> 
> https://github.com/paritytech/substrate-archive/blob/6e5a6e2d22d7dd28bfb4f2eb50e5f2de6045e433/substrate-archive/src/archive.rs#L350
> 
> and then replace `config.control.task_url` from there if it exists. this way env variable always takes precedence over the config file

joaoepj:
Fixing code style.

PR #462, opened 2022-04-29
optimize extrinsics decoding and storage indexing

xcaptain:
1. Ignore extrinsics decoding errors and move forward rather than abort and retry.
2. Index storage in ascending order and with a fixed limit(1000) to avoid too many blocks not in storage table.
3. Construct wasm executor from runtime configs rather than hard-coded values.

### DIFF
--- a/substrate-archive-backend/src/read_only_backend.rs
+++ b/substrate-archive-backend/src/read_only_backend.rs
@@ -140,16 +140,13 @@ where
 				if key.len() == 4 && fun(num) {
 					let head: Option<Block::Header> = readable_db
 						.get(super::util::columns::HEADER, &value)
-						.map(|bytes| Decode::decode(&mut &bytes[..]).ok())
-						.flatten();
+						.and_then(|bytes| Decode::decode(&mut &bytes[..]).ok());
 					let body: Option<Vec<Block::Extrinsic>> = readable_db
 						.get(super::util::columns::BODY, &value)
-						.map(|bytes| Decode::decode(&mut &bytes[..]).ok())
-						.flatten();
+						.and_then(|bytes| Decode::decode(&mut &bytes[..]).ok());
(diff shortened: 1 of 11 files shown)


insipx:
This looks good to me, some simple and effective changes

xcaptain:
Added a small improvement, please take a look again.

insipx (approved):

jsdw (approved):

jsdw:
I think you'll need to run/resolve `cargo fmt` and `cargo clippy`. I'm not sure what's going on with the tests offhand!

xcaptain:
The new commit may fix the CI error. The reason is in previous commits I changed behavior of `missing_storage_blocks`, add one more condition: https://github.com/paritytech/substrate-archive/pull/462/files#diff-c1534283c28a073545083cd49a8459f91e52899eb9289b7098327c814d061dc7R199

xcaptain:
Considered twice, it's better not to add the `AND block_num >= (SELECT MAX()...)` condition. Adding it won't make the sql execution faster.
@insipx @jsdw please take a look again.

xcaptain:
Accepted some clippy fixes, please run CI again.

insipx:
LGTM. I would just make sure that if archive stops unexpectedly or is stopped, then we don't miss any storage blocks since we specify that we restore at MAX(block_num).

As long as we always have contiguous blocks in `storage` it's ok to index from `MAX(block_num) in storage`. Have you ran archive to test & make sure that this holds?

insipx:
If you're running with only one storage indexer than I believe you are correct. If a later block finishes execution before an earlier block and is inserted first, then a gap is created which `restore_missing_blocks` would miss if indexing from `MAX(block_num)`

insipx (approved):

